PR #720, opened 2022-02-15
fix: Remove the blank line from the contact information in text output

### LINKED ISSUE #700: format of <contact>
`type_enhancement`   |    by sginoza@amsl.com

___

Per discussion on tools-discuss, remove the blank line that appears before email in the Contributors and Authors' Addresses sections.  The blank line only appears in the text output.  This update results in a consistent format in those sections and across outputs.

Note that this only applies to Contributors when it has the same format as the Authors' Addresses.

See the thread here:
https://mailarchive.ietf.org/arch/msg/tools-discuss/5HbO_EjBpNvbchBpZqvBbaduEdM/

___
_Issue migrated from trac:700 at 2022-02-08 07:18:53 +0000_

kesara:
Fixes #700


rjsparks (approved):
Splitting the commits this way made what happened much clearer.

PR #725, opened 2022-02-17
fix: build PDF with Weasyprint 53

dkg:
This patch is from @kitterma, i'm just forwarding it from https://trac.ietf.org/trac/xml2rfc/ticket/696 (it's also at https://bugs.debian.org/1001824)

Weasyprint 53 and later no longer use cairo and pango, so xml2rfc tests for PDF generation capability need updating.  Also, it appears that test_text_ content is obsolete with the new Weasyprint.  The test fails even though PDF generation succeeds.

On debian and debian-derived systems, the suggestions that remain in run.py when weasyprint isn't available are rather over-complicated,but debian can continue to patch in more streamlined suggestions for its distribution.

### DIFF
--- a/test.py
+++ b/test.py
@@ -499,18 +499,18 @@ class PdfWriterTests(unittest.TestCase):
         self.pdfxml = copy.deepcopy(self.elements_pdfxml)
         self.root = copy.deepcopy(self.elements_root)
 
-    def test_text_content(self):
-        def norm(t):
-            return re.sub(r'\s+', ' ', t).strip()
-        #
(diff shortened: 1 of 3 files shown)


kesara:
@dkg, @kitterma, Thank you for the PR and apologies for taking long to reply.
Because this change can affect the PDF output, we will have to regenerate a few RFCs to compare and discuss with RPC.
I'll arrange this.

kesara:
Closing this PR in favour of #802.
weasyprint 53+ changes required further changes to the test suite.
Thank you for your support.

Issue #722, opened 2022-02-17
--table-borders should be removed?


miekg:
--table-borders changes the way a document looks and should be removed or moved into the <rfc> tag.

I *think* most folks are/were in favor of just removing it.

cabo:
> I _think_ most folks are/were in favor of just removing it.

The command line option is somewhat nonsensical, indee.

Not at all the formatting parameter.
Most usefully as an attribute.
I wouldn't mind a PI either.

miekg:
attribute is a nice idea.

Also in similar vein: `--table-hyphen-breaks` should be made a default and not a flag?

cabo:
These are what we call "tweaks" in the world of document formatting.
It does not make sense to apply tweaks globally for a document, unless that is so uniform that all (e.g.) tables are the same.  Instead, the tweaks should be with the specific item.  PIs were meant to be one channel for tweaks (they can also be used to indicate whether the RFC should be stapled after printing, but that is not something we embrace).  We do have several tweaks that are attributes, and nothing is wrong adding a few more, if we don't go overboard.

cabo:
I see a lot of people using ASCII artwork because they don't get the table rendering they need.
That of course can impair the HTML/PDF rendering while improving the TXT rendering.

dkg:
It would be great to have the table borders in the text output rendered with Unicode U+2500 BOX DRAWING characters instead of ASCII art, and i agree that it doesn't make a lot of sense as a command-line option.

The directive should be able to be placed either be at the document level, or as an option for specific tables to override the document-level preference.

jrlevine:
This gets into the swamp of what non-ASCII characters we allow in the text rendering.
I agree that the RFC 7997 turned out to be too strict but I really do not want to undo this one casual hack at a time.

In the meantime, those HTML tables look pretty nice.

miekg:
> In the meantime, those HTML tables look pretty nice.

yep. But call me old or whatever, but I'm still very charmed by the xml2rfc text output.

cabo:
So maybe  we should starting labeling issues like this post-7997 (i.e., revisit when we have managed to enter a state where we can discuss 7997).

miekg:
sgtm - none of this seems urgent

PR #727, opened 2022-02-18
ci: support python 3.10

dkg:
This indicates that xml2rfc should work fine with python 3.10 as well
as the earlier versions.

I haven't tested (nor do i fully understand) the tox.ini and
.github/workflows/checks.yml changes, but hopefully some automated CI
will run against them and report if there are any problems.

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -90,7 +90,7 @@ jobs:
         echo "Installing pip + wheel..."
         python -m pip install --upgrade pip wheel
         echo "Installing requirements.txt + test dependencies..."
-        python -m pip install -r requirements.txt tox tox-gh-actions certifi decorator dict2xml==1.6 "pycairo<1.20" PyPDF2 "weasyprint<53"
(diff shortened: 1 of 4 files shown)


dkg:
The problem is that it's working with version 1.6 of the `dict2xml` python module:

https://github.com/ietf-tools/xml2rfc/runs/5241008281?check_suite_focus=true#step:7:212

See [https://pypi.org/project/dict2xml/1.7.0/](the `dict2xml` changelog for 1.7.0, which reads):

> Use collections.abc to avoid deprecation warning. Thanks @mangin.

kesara:
The latest problem is caused by `makupsafe`: https://markupsafe.palletsprojects.com/en/2.1.x/changes/?highlight=soft_unicode#version-2-1-0
This `makeupsafe` is a dependency of `jinja2` maybe removing `jinja2` restriction might help this.
#661 addressed this but change is not in the main branch yet.

kesara:
Closing this PR in favour of #772.
#722 has cherry picked changes from this PR.

PR #760, opened 2022-04-21
fix: Don't attempt to select initials when fullname contains non Latin characters

### LINKED ISSUE #759: Don't attempt to select initials when fullname contains non Latin and non Latin-1 supplement characters
(Breakaway from: #750)

Don't attempt to select initials when `fullname` contains non [basic latin](https://www.unicode.org/charts/PDF/U0000.pdf) and non [latin-1 supplement](https://www.unicode.org/charts/PDF/U0080.pdf) characters.

Input:
```
<author asciiFullname="Kazuho Oku" fullname="奥 一穂">
```

Current output:
```
奥. 一穂 (K. Oku)
```

Desired output:
```
奥 一穂 (K. Oku)
```

Note that `fullname="Patrik Fältström"` should not trigger the display of `fullname`.

kesara:
Fixes #759

### DIFF
--- a/tests/input/draft-v3-features.xml
+++ b/tests/input/draft-v3-features.xml
@@ -32,6 +32,8 @@
         </postal>
       </address>
     </author>
+    <author fullname="Armin Bergström">
+    </author>
     <date year="2017" month="04" day="15"/>
     <abstract>
       <t>
--- a/xml2rfc/util/name.py
+++ b/xml2rfc/util/name.py
@@ -34,11 +34,14 @@ def short_author_name_parts(a):
     else:
         fullname = a.get('fullname') or ''
         if fullname:
-            if len(fullname.split())>1:
-                parts = fullname.split()
-                initials = ' '.join([ "%s."%n[0].upper() for n in parts[:-1] ])
-                surname  = parts[-1]
-                parts = [initials, surname ]
+            if is_script(fullname, 'Latin'):
+                if len(fullname.split())>1:
+                    parts = fullname.split()
+                    initials = ' '.join([ "%s."%n[0].upper() for n in parts[:-1] ])
+                    surname  = parts[-1]
+                    parts = [initials, surname ]
+                else:
+                    parts = [ None, fullname ]
             else:
                 parts = [ None, fullname ]
         else:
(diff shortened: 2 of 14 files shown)


xml2rfc/util/name.py
```diff
@@ -34,11 +34,14 @@ def short_author_name_parts(a):
     else:
         fullname = a.get('fullname') or ''
         if fullname:
-            if len(fullname.split())>1:
-                parts = fullname.split()
-                initials = ' '.join([ "%s."%n[0].upper() for n in parts[:-1] ])
-                surname  = parts[-1]
-                parts = [initials, surname ]
+            if is_script(fullname, 'Latin'):
```
cabo, line 37: Maybe this isn't important, as we are pivoting away from the use of initials, but I'm not sure that a Latin writing system is a necessary indicator for the use of basic initials heuristics.  E.g., Cyrillic or Greek use the same initials heuristics (I think), while in general that is of course not true.  So I think I'd like to have some documentation what exactly this change does.

rjsparks: I'm not sure where we would put such documentation - ultimately this is something the style guide will describe and may that's the documentation? From an implementation point of view, this is a rats-nest, and this fix is showing to get us what we need for the short term (an RFC containing a Japanese name is on the threshold of release). I suggest we approach this with incremental changes as real cases arrive.

cabo: Works for me.

There should be something that I can tell kramdown-rfc users about what these attributes mean and how they can be successfully used; this is what I meant with "documentation".  That information would probably live in Github issues for quite a while.

tests/input/draft-v3-features.xml
```diff
@@ -32,6 +32,8 @@
         </postal>
       </address>
     </author>
+    <author asciiFullname="Armin Bergstrom" fullname="Armin Bergström">
```
cabo, line 35: This example should never happen.  "asciiFullname" is not about ASCII at all.  Armin's name is fully Latin so doesn't need an asciiFullname.

tests/valid/rfc6787.exp.xml
```diff
@@ -11338,7 +11338,7 @@ identification-tag =    token
 
       <!--Internet Message Format-->
 
-      <reference anchor="RFC5322" target="https://www.rfc-editor.org/info/rfc5322" xml:base="https://www.rfc-editor.org/refs/bibxml/reference.RFC.5322.xml" quote-title="true">
+      <reference anchor="RFC5322" target="https://www.rfc-editor.org/info/rfc5322" xml:base="https://xml2rfc.tools.ietf.org/public/rfc/bibxml/reference.RFC.5322.xml" quote-title="true">
```
cabo, line 11341: How does xml2rfc.tools.ietf.org enter the picture again?

tests/input/draft-v3-features.xml
kesara: Thanks, I'll correct this.

tests/valid/rfc6787.exp.xml
kesara: This is coming from `tests/cache` in my local machine. 🤦

rjsparks (approved):

PR #764, opened 2022-04-27
fix: Make long sourcecode sections breakable

### LINKED ISSUE #739: PDFs have blank pages before sourcecode or artwork
If the first thing in a section is `sourcecode` or `artwork`, each preceding section header is on a separate page.
Possible fixes:

Option 1. Remove CSS rule that avoid inside breaks for `sourcecode`.

Option 2. xml2rfc can add a class that allows `sourcecode` to break inside when `sourcecode` is larger than 80 lines or
so.

Option 3. Schema change. We can introduce a new attribute to `sourcecode` element to tell it's okay to break inside. So
xml2rfc can use a different class for that.

Option 4: render a dummy blank paragraph before the `sourcecode`

kesara:
fixes #739

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -61,7 +61,7 @@ pyfiles  = $(wildcard  xml2rfc/*.py) $(wildcard  xml2rfc/writers/*.py)
 tests: minify tests-no-network cachetest
 
 # Tests that can run without network access
-tests-no-network: test flaketest drafttest rfctest utf8test v3featuretest elementstest indextest bomtest wiptest mantest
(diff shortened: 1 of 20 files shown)


rjsparks (approved):

rjsparks:
Are we just relying on visual inspection to verify the change in the pdf?

kesara:
> Are we just relying on visual inspection to verify the change in the pdf?

There's a test for this see `tests/input/sourcecode.xml`.
Also, there are PDFs generated for visual inspection in a test server.

PR #766, opened 2022-04-28
fix: Render text without toc

### LINKED ISSUE #765: TXT generation fails while HTML and PDF succeed, error message is unhelpful
The attached XML fails to render as txt, with the unhelpful error message "xml2rfc error: text generation error". PDF and HTML succeed. I accept that there might be something wrong with the XML (which is quite old), but the error message doesn't help me figure out what it is.

For that matter, it would be nice to be able to see the more verbose errors-and-warnings form the old xml2rfc site offered.

P.S.: GitHub refused XML attachment, so I added a TXT extension to the attachment to circumvent that
[test.xml.txt](https://github.com/ietf-tools/ietf-at-ui/files/8584931/test.xml.txt)
.

cabo:
(Add tests and shake.)


rjsparks (approved):

kesara:
Fixes #765

PR #771, opened 2022-05-09
fix: Display ASCII names for authors in references 

### LINKED ISSUE #768: Fix non latin name rendering references in text output
(Breakaway from: https://github.com/ietf-tools/xml2rfc/issues/750)

XML:
```
        <reference anchor="RFC9218" target="https://www.rfc-editor.org/info/rfc9218">
          <front>
            <title>Extensible Prioritization Scheme for HTTP</title>
            <author asciiFullname="Kazuho Oku" fullname="奥 一穂"/>
            <author initials="L" surname="Pardue" fullname="Lucas Pardue"/>
            <date year="2022" month="March"/>
          </front>
          <seriesInfo name="RFC" value="9218"/>
          <seriesInfo name="DOI" value="10.17487/RFC9218"/>
        </reference>
      </references>
```

Text output:
```
              奥 一穂 and L. Pardue, "Extensible Prioritization Scheme for
              HTTP", RFC 9218, DOI 10.17487/RFC9218, March 2022,
              <https://www.rfc-editor.org/info/rfc9218>.
```

HTML/PDF outputs:
```
奥 一穂 (Oku, K.) and L. Pardue, "Extensible Prioritization Scheme for HTTP", RFC 9218, DOI 10.17487/RFC9218, March 2022, <https://www.rfc-editor.org/info/rfc9218>.
```

Text output should be similar to HTML/PDF outputs and display ASCII name.

kesara:
Display ASCII names for authors in references with non-Latin characters in
their names.

Fixes #768

### DIFF
--- a/xml2rfc/writers/text.py
+++ b/xml2rfc/writers/text.py
@@ -28,8 +28,9 @@ from xml2rfc.writers.base import default_options, BaseV3Writer, RfcWriterError
 from xml2rfc import utils
 from xml2rfc.uniscripts import is_script
 from xml2rfc.util.date import extract_date, augment_date, get_expiry_date, format_date
-from xml2rfc.util.name import short_author_name, short_author_ascii_name, short_author_name_parts, short_org_name_set
-
+from xml2rfc.util.name import (short_author_name, short_author_ascii_name,
+                               short_org_name_set, ref_author_name_first,
+                               ref_author_name_last)
 from xml2rfc.util.name import full_author_name_set
 from xml2rfc.util.num import ol_style_formatter, num_width
 from xml2rfc.util.unicode import expand_unicode_element, textwidth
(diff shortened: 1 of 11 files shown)


xml2rfc/writers/text.py
```diff
@@ -1125,28 +1126,21 @@ def render_authors(self, e, width, **kwargs):
...
-                                     surname)
-                    else:
-                        buf.append(surname)
-                elif len(initials) > 0:
-                    buf.append(surname + ', ' + initials)
-                else:
-                    buf.append(surname)
-                if author.attrib.get('role', '') == 'editor':
-                    buf.append(', Ed.')
-            elif organization is not None and organization.text:
```
cabo, line 1145: So we no longer the organization if the author name is missing?  I think that is a regression.

kesara: [`ref_author_name_last`](https://github.com/ietf-tools/xml2rfc/blob/6e64a1c667892a9600c2a772615c1be3df855936/xml2rfc/util/name.py#L87) and [` ref_author_name_last`](https://github.com/ietf-tools/xml2rfc/blob/6e64a1c667892a9600c2a772615c1be3df855936/xml2rfc/util/name.py#L112) adds the `organization` if `name` is missing, so no longer needed to include that check.
Both HTML and PDF renderings make use of the same functions.

cabo: Oh, this is done un util/name.py (ref_author_name_first/last).

kesara: Yeah.

cabo: A differences that ref_.... uses short_org...., while the existing code used the full name of the org.

cabo: It appears that was a difference between text and html.  That is now fixed!?

JayDaley: I have asked the RPC to advise

alicerusso: This seems to be about how an organization's name appears in a reference when that org is an author. I think short name (when present) is preferred for the following reasons:
  * readability of the reference itself (assuming the org is well known by the short name)
  * usage in new bibxml (e.g., reference.IEEE 802-1AX.2020.xml) includes the abbrev attribute and it's preferred
```<organization abbrev="IEEE">Institute of Electrical and Electronics Engineers</organization>```

Whether or not to use an abbreviated name for orgs in references is not included in current style guidance. Over time, either has been used. FWIW, the RPC often uses only the abbreviated name. For example: rfc9182.notprepped.xml contains
```
<author><organization>ISO</organization></author> for [ISO10589]
<author><organization>IEEE</organization></author> for [IEEE802.1AX]
```

cabo: Thank you!
So it seems the PR as proposed supports this approach best, while harmonizing TXT and HTML/PDF renderings.

rjsparks (approved):

PR #772, opened 2022-05-10
build: Add support for Python 3.10

kesara:
This PR consists of cherry picked changes from #727 because the original PR is stale and complicated to update via GitHub.

### DIFF
--- a/xml2rfc/writers/doc.py
+++ b/xml2rfc/writers/doc.py
@@ -149,7 +149,8 @@ class DocWriter(base.BaseV3Writer):
             # This depend on at least one generic option with title being listed before '--config-file'
             if a.container.title == 'Generic Options with Values':
                 genargs_group = a.container
-            if a.container.title == 'optional arguments':
+            if a.container.title == 'optional arguments' or a.container.title == 'options':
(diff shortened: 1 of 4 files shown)


xml2rfc/writers/doc.py
```diff
@@ -149,7 +149,8 @@ def process(self):
             # This depend on at least one generic option with title being listed before '--config-file'
             if a.container.title == 'Generic Options with Values':
                 genargs_group = a.container
-            if a.container.title == 'optional arguments':
+            if a.container.title == 'optional arguments' or a.container.title == 'options':
+                # Python 3.10 calls this options while Python 3.6 to Python 3.8 calls this optional arguments
```
rjsparks, line 153: what does 3.9 do?

rjsparks (approved):

xml2rfc/writers/doc.py
kesara: This was changed in Python 3.10: https://docs.python.org/3.10/whatsnew/3.10.html#argparse
Sorry, I missed this before the merge.

kesara: I have updated the comment.

Issue #769, opened 2022-05-06
Crashes when year is no numeric


kesara:
`xml2rfc` crashes when `date` element's `year` attribute contains non numeric characters.
Example:
```
<date year="2016  "/>
```

`xml2rfc` must emit an error message instead of crashing.

kesara:
Fixed with #795

PR #795, opened 2022-06-12
fix: Handle date type errors gracefully

### LINKED ISSUE #794: Crash on <date day="??"
```
            <date year="2017" month="September" day="??"/>
```

in a reference is enough to crash xml2rfc:

```
Traceback (most recent call last):
  File "/Users/cabo/bin/xml2rfc", line 8, in <module>
    sys.exit(main())
  File "/Users/cabo/lib/python3.9/site-packages/xml2rfc/run.py", line 737, in main
    xmlrfc.tree = prep.prep()
  File "/Users/cabo/lib/python3.9/site-packages/xml2rfc/writers/preptool.py", line 216, in prep
    tree = self.dispatch(self.selectors)
  File "/Users/cabo/lib/python3.9/site-packages/xml2rfc/writers/base.py", line 1925, in dispatch
    func(e, e.getparent())
  File "/Users/cabo/lib/python3.9/site-packages/xml2rfc/writers/preptool.py", line 419, in check_attribute_values
    if i and not i.isdigit() and not i==self.get_attribute_defaults(c.tag)[a]:
KeyError: 'day'
```

cabo:
(There may be a better solution, but this fixes the crash for now.)

### DIFF
--- a/xml2rfc/writers/preptool.py
+++ b/xml2rfc/writers/preptool.py
@@ -661,7 +661,10 @@ class PrepToolWriter(BaseV3Writer):
                 month = today.strftime('%m')
                 day = today.strftime('%d')
             datestr = "%s-%s-%s" %(year, month, day or '01')
-            date = datetime.datetime.strptime(datestr, "%Y-%m-%d").date()
+            try:
+                date = datetime.datetime.strptime(datestr, "%Y-%m-%d").date()
+            except ValueError:
+                self.die(e, "<date> {} is invalid'".format(datestr))
             if abs(date - datetime.date.today()) > datetime.timedelta(days=3):
                 self.warn(d, "The document date (%s) is more than 3 days away from today's date" % date)
             n = self.element('date', year=year, month=month, line=d.sourceline)
--- a/xml2rfc/writers/preptool.py
+++ b/xml2rfc/writers/preptool.py
@@ -416,7 +416,7 @@ class PrepToolWriter(BaseV3Writer):
         for c in e.iter(tags):
             for a in integer_attributes[c.tag]:
                 i = c.get(a)
-                if i and not i.isdigit() and not i==self.get_attribute_defaults(c.tag)[a]:
+                if i and not i.isdigit() and not i==self.get_attribute_defaults(c.tag).get(a):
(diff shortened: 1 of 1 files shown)


xml2rfc/writers/preptool.py
```diff
@@ -416,7 +416,7 @@ def check_attribute_values(self, e, p):
         for c in e.iter(tags):
             for a in integer_attributes[c.tag]:
                 i = c.get(a)
-                if i and not i.isdigit() and not i==self.get_attribute_defaults(c.tag)[a]:
+                if i and not i.isdigit() and not i==self.get_attribute_defaults(c.tag).get(a):
                     self.err(c, 'Expected <%s> attribute "%s" to be a non-negative integer, but found "%s"' % (c.tag, a, i))
```
kesara, line 420: I think, it's better to use `self.die()` instead of `self.err()` to exit the processing.
With the current change `xml2rfc` crashes again because it can't handle the date conversion.

cabo: Works for me.
My main objective was to get out the error message before croaking.

cabo: Actually, this is a generic loop, not just for dates, and it would help to get all error messages out.

kesara: Yeah, this might result in unwanted failures.
I think it's better to handle this at the date conversion.
I can push a change.

xml2rfc/writers/preptool.py
```diff
@@ -661,7 +661,10 @@ def front_insert_date(self, e, p):
                 month = today.strftime('%m')
                 day = today.strftime('%d')
             datestr = "%s-%s-%s" %(year, month, day or '01')
-            date = datetime.datetime.strptime(datestr, "%Y-%m-%d").date()
+            try:
+                date = datetime.datetime.strptime(datestr, "%Y-%m-%d").date()
+            except ValueError:
+                self.die(e, "<date> {} is invalid'".format(datestr))
```
kesara, line 667: @cabo, this will catch any date conversion issues and die with an error.

kesara:
This fixes #769 as well.

cabo:
I can't approve changes to my own PR, but this does look good (did not test it, though).

rjsparks (approved):

PR #802, opened 2022-06-28
feat: Update weasyprint to 55.0 

### LINKED ISSUE #696: Updates For Weasyprint > 53
`type_defect`   |    by scott@kitterman.com

___

As I gather you know already, since the test dependency on weasyprint is pinned to versions < 53, as released, xml2rfc does not work with newer versions of weasyprint.  I've done some testing and with some adjustments to the checks for PDF capability, it appears to work OK.

With the patch that I will hopefully be able to attach later I was able to take a recent XML file for an ID and successfully create a PDF format document, but there are still tests that fail:

a/xml2rfc-3.12.0/test.py", line 513, in test_included_fonts
    if xml2rfc.HAVE_WEASYPRINT and xml2rfc.HAVE_PYCAIRO and xml2rfc.HAVE_CAIRO and xml2rfc.HAVE_PANGO:

This will trivially need updating since HAVE_CAIRO and HAVE_PANGO no longer apply.  If those are present, the test passes.

======================================================================
FAIL: test_text_content (__main__.PdfWriterTests)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/xml2rfc-3.12.0/test.py", line 510, in test_text_content
    self.assertIn(t, text)
AssertionError: 'RFC' not found in 'l- WN –&,ÛžÆ TNDQ&·sX l-%Zñ¢<³ WN P½RÔ n®>Ø %˜oh l† .µ ¬( .µ=N'

----------------------------------------------------------------------

This is more problematic.  Despite some effort, I'm unable to figure this one out.  I wonder though, with the new weasyprint that uses pydyf, is this even a valid test, since generating PDFs does work and if this were testing anything important/relevant it seems like that wouldn't be possible?

We updated weasyprint in Debian without realizing the problems this would cause, so it would be really nice to get this sorted.  

___
_Issue migrated from trac:696 at 2022-02-08 07:18:40 +0000_

kesara:
* Update weasyprint to 55.0.
* Replace PyPDF2 with pdfplumber.

This changes PDF generation from cairo to pydyf.

Fixes #696

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -71,11 +71,10 @@ COPY requirements.txt .
 
 # build xml2rfc
 RUN pip3 install -r requirements.txt \
-    "weasyprint==55.0" \
+    "weasyprint>=53.0" \
     decorator \
     dict2xml \
-    typing-extensions \
-    pdfplumber
+    "pypdf2>=2.6.0"
 RUN make install
 
 # cleanup
@@ -84,7 +83,7 @@ RUN rm roboto-mono.zip
 RUN apt-get remove --purge -y software-properties-common wget unzip
 RUN apt-get autoclean
 RUN apt-get clean
-RUN pip3 uninstall -y decorator dict2xml typing-extensions pdfplumber
+RUN pip3 uninstall -y decorator dict2xml pypdf2
 RUN rm setup.py README.md Makefile configtest.py requirements.txt
 RUN rm -r xml2rfc build dist
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,4 @@
-FROM ubuntu:focal
-# deadsnakes has Python3.6 for focal but not jammy
+FROM ubuntu:latest
 LABEL maintainer="Kesara Rathnayake <kesara@staff.ietf.org>"
 
 ENV DEBIAN_FRONTEND noninteractive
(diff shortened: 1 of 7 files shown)


dkg:
fwiw, this will make things harder for me to maintain and adequately test in debian, since debian doesn't have pdfplumber.  I've just gotten [pypdf2 fixed upstream](https://github.com/py-pdf/PyPDF2/pull/1118) to be able to correctly handle the types of output produced by more recent versions of weasyprint.

For simplicity and ease of maintenance and testing, my preference would be:

- avoid the switch from pypdf2 to pdfplumber for testing, just require pypdf2 version 2.6.0 and above (first version with the fix to avoid the CMap crashing)
- drop all explicit mentions of cairo
- require the first version of weasyprint that dropped cairo (53.0b1) or any newer one

kesara:
@dkg, Thanks for the update.
pypdf2 version 2.6.0 seems to fix the issues that I had with PDF test, so I can drop that commit.
Since WeasyPrint 53.0b1 is a pre-release, I'd like to set v53.0 as the minimum Weasyprint requirement. Hope that's okay.

Dockerfile
```diff
@@ -70,11 +71,10 @@ COPY requirements.txt .
 
 # build xml2rfc
 RUN pip3 install -r requirements.txt \
-    "weasyprint==52.5" \
-    "pycairo<1.20" \
+    "weasyprint>=53.0" \
```
rjsparks, line 74: Shouldn't this be >=55 now?

rjsparks (approved):

Dockerfile
dkg: Why >= 55?  weasyprint is actually at 56 now upstream, but there's no good reason to increase the version if you don't actually need the newer version.

Put another way: please do not gratuitously increase the minimum version needed.  if 53.0 works as well, there's no reason to have the elevated version requirements.

If there is a specific reason that 55 is necessary, then of course make it required, but otherwise, you do the ecosystem a favor by indicating the flexibility explicitly.

dkg:
@kesara wrote:

> Since WeasyPrint 53.0b1 is a pre-release, I'd like to set v53.0 as the minimum Weasyprint requirement. Hope that's okay.

Yes, that makes sense to me.  I was being hyper-precise by indicating the first visible release that dropped cairo, but you're right that it makes sense to only target "official" release versions.

xml2rfc/run.py
```diff
@@ -84,29 +80,14 @@ def get_pdf_help(missing_libs=""):
...
-    
-        PYTHONPATH=/Users/username/Library/Python/2.7/lib/python/site-packages
-
-    for Python 2.7.
-
-    The basic pip commands (modify as needed according to the text above)
-    are:
-
-        pip install 'pycairo>=1.18' 'weasyprint<=0.42.3'
+        pip install 'weasyprint==55.0'
```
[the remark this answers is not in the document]
dkg: should be >=53.0, probably.

Dockerfile
kesara: I don't see any specific reason for using 55 or 56 at this point.
So I'm happy to keep `weasyprint>=53.0` wider compatibility.
But services like author tools will start using the latest available version.

dkg: yep, that sounds entirely reasonable.  the specified version ranges are constraints, but of course any specific instance will choose whatever version of the package it wants as long as it's within the constraints.

PR #839, opened 2022-07-23
Move sourcecode classes

martinthomson:
The HTML elements created for a `<sourcecode>` element have class
attributes attached to the `<pre>` element.  In comparison, `<artwork>`
is rendered with class attritributes attached to the outer `<div>`.

Having the classes on the outer `<div>` makes it a LOT easier to style
these properly, so I've moved them in this CL.

### DIFF
--- a/xml2rfc/data/xml2rfc.css
+++ b/xml2rfc/data/xml2rfc.css
@@ -791,7 +791,7 @@ tt, code, pre, code {
 }
 
 /* Fix the height/width aspect for ascii art*/
-pre.sourcecode,
+.sourcecode pre,
 .art-text pre {
   line-height: 1.12;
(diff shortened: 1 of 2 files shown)


kesara:
This change requires updates to `xml2rfc.css` because there are `pre.<class>` specific rules.
Example:
https://github.com/ietf-tools/xml2rfc/blob/main/xml2rfc/data/xml2rfc.css#L795

martinthomson:
@kesara, thanks for pointing that out.  I think that I've fixed it.

kesara (changes requested):
`pre.breakable` should be changed to `.breakable pre`.

kesara (approved):

PR #842, opened 2022-07-24
 fix: Prevent submission date warnings for RFCs 

### LINKED ISSUE #734: Warning about dates isn't appropriate for an RFC
```
.../quic/rfc9000.xml(19): Warning: The document date (2021-05-01) is more than 3 days away from today's date
```

This warning is fine in cases where internet-drafts are being prepared for submission, but it makes no sense for an RFC where the publication date is immutable.

kesara:
Fixes #734


kesara:
@rjsparks I confirmed with RPC that they are not using this as a safety check & they are okay with the change.

rjsparks (approved):

PR #847, opened 2022-07-28
Add xml2rfc class to HTML body element

martinthomson:
It is sometimes desirable to take the HTML generated by xml2rfc and embed it in another web page.  In that context, being able to reuse existing styling is also desirable.  This change doesn't do anything to support this process other than include a class attribute on `<body>`.

Currently, a lot of the style rules target the `<body>` element, largely by necessity.  With a class attribute, it becomes possible to target that element with the same style rules, even when the content of the page is embedded in another context as a `<div>` or other element.


kesara (approved):

kesara:
@martinthomson, I can not push the updated commits with conflicts resolved to this PR. I have a branch with these conflicts fixed here: https://github.com/kesara/xml2rfc/tree/xml2rfc-class

Do you like to update your PR with commits off that (or create a new PR from that branch).

_I can create a PR too, but that results in wrong release notes / change log giving credits to me rather than you._

kesara:
> @martinthomson, I can not push the updated commits with conflicts resolved to this PR. I have a branch with these conflicts fixed here: https://github.com/kesara/xml2rfc/tree/xml2rfc-class
> 
> Do you like to update your PR with commits off that (or create a new PR from that branch).
> 
> _I can create a PR too, but that results in wrong release notes / change log giving credits to me rather than you._

I found a way to deal with this issue by changing my workflow to not do automatic rebase. :)

PR #849, opened 2022-08-01
build: Update xml2rfc docker image

### LINKED ISSUE #846: Docker image performance
Using run.sh is a useful way of doing development.  It's annoying that the image doesn't have git and a few basic tools, but once you get around that, the requirements are easily managed.

I do find it very frustrating that the build basically always runs the pip install from requirements.txt.  If the steps were reordered so that copying requirements.txt and installing from it came *before* a lot of the other copy commands, then it would be much faster to run in common cases.

kesara:
* build: Change docker publish workflow to run manually [skip ci]
* build: Install xml2rfc from PyPI [skip ci]
* build: Add additional tools [skip ci]

Fixes #846


Dockerfile
```diff
@@ -61,35 +60,26 @@ RUN unzip -q roboto-mono.zip -d ~/.fonts/opentype/
...
 
-# build xml2rfc
-RUN pip3 install -r requirements.txt \
-    "weasyprint>=53.0" \
-    decorator \
-    dict2xml \
-    "pypdf2>=2.6.0"
-RUN make install
+# Install latest packages from PyPI
+RUN pip3 install xml2rfc "weasyprint>=53.0" tox
```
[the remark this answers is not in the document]
kesara: I was treating this as a hybrid image for both development and anyone who wants to use xml2rfc with all required dependencies and fonts.

But as @NGPixel, mentioned there's a separate base image which I wasn't aware of.
So I'll update to make use of the base image.
And include all the dependencies required to build the xml2rfc in this dev image.

rjsparks (approved):

docker/dev.Dockerfile
```diff
@@ -0,0 +1,50 @@
...
+RUN apt-get update --fix-missing
+
+# Install additional tools
+RUN apt-get install -y --fix-missing \
+    git \
+    vim \
+    less
+
+# Set locale
+RUN locale-gen en_US.UTF-8
```
NGPixel, line 18: Already set in base image

NGPixel (approved):

docker/base.Dockerfile
```diff
@@ -1,66 +1,76 @@
...
+RUN ln -sf ~/.fonts/opentype/*.[to]tf /usr/share/fonts/truetype/
+RUN fc-cache -f
+
+# Copy everything required to build xml2rfc
+COPY setup.py .
+COPY README.md .
+COPY LICENSE .
+COPY Makefile .
+COPY configtest.py .
+COPY xml2rfc ./xml2rfc
```
martinthomson, line 58: These steps, particularly the last one, involve files that will change frequently.  If your goal is to improve layer caching, then moving those steps to the end will improve that greatly.

docker/base.Dockerfile
```diff
@@ -1,66 +1,76 @@
...
+RUN make install
+
+# cleanup
+RUN rm Noto-unhinted.zip
+RUN rm roboto-mono.zip
+RUN apt-get remove --purge -y software-properties-common wget unzip
+RUN apt-get autoclean
+RUN apt-get clean
+RUN pip3 uninstall -y decorator dict2xml pypdf2
+RUN rm setup.py Makefile configtest.py requirements.txt
```
martinthomson, line 70: Running these as discrete steps won't reduce the size of the docker image, it only increases it. I'd cut these.

I'm interpreting this docker image as including a test, which is not really something that a dockerfile is well suited toward.  If you want to test, maybe you can run these steps in the workflow instead.

docker/dev.Dockerfile
```diff
@@ -0,0 +1,47 @@
...
+RUN pip3 install \
+    tox \
+    decorator \
+    dict2xml \
+    "pypdf2>=2.6.0"
+
+# cleanup
+RUN apt-get remove --purge -y software-properties-common
+RUN apt-get autoclean
+RUN apt-get clean
```
martinthomson, line 42: To save space, cleanup should be run in the same step as the files that are being created.

Also, I wouldn't bother removing software-properties-common; it's just not that much space to save.

docker/base.Dockerfile
```diff
@@ -1,66 +1,76 @@
...
+RUN pip3 install -r requirements.txt \
+    "weasyprint>=53.0" \
+    decorator \
+    dict2xml \
+    "pypdf2>=2.6.0"
+
+# Install required fonts
+RUN mkdir -p ~/.fonts/opentype
+RUN wget -q https://noto-website-2.storage.googleapis.com/pkgs/Noto-unhinted.zip
+RUN unzip -q Noto-unhinted.zip -d ~/.fonts/opentype/
```
martinthomson, line 46: Running these as a single RUN step will reduce the number and size of the layers that are created.

docker/base.Dockerfile
```diff
@@ -1,66 +1,76 @@
...
+RUN wget -q https://fonts.google.com/download?family=Roboto%20Mono -O roboto-mono.zip
+RUN unzip -q roboto-mono.zip -d ~/.fonts/opentype/
+RUN ln -sf ~/.fonts/opentype/*.[to]tf /usr/share/fonts/truetype/
+RUN fc-cache -f
```
martinthomson, line 50: Same here.

kesara:
@martinthomson, There's a workflow now that builds `xml2rfc-dev` image: https://github.com/ietf-tools/xml2rfc/pkgs/container/xml2rfc-dev

I have updated `./run` to make use of that.
So anyone with amd64 or arm64 will not have to build the `xml2rfc-dev` image anymore.

martinthomson:
Thanks @kesara, that should work out to be a lot faster.  I made a few suggestions that might help with image size, for those of us who live on the wrong side of the planet.

kesara:
@martinthomson, Thanks for the suggestions and I'm looking to implement them.
I need to learn more about how docker layers get created.

martinthomson:
Think of each command in the dockerfile as making some changes on top of the previous state.  Each change needs to be saved and adds to the size of the final image.  If you create smaller changes or fewer changes, the final image can be a lot smaller.

This means consolidating actions into a single step is good.  It also means that if you do something that creates temporary files (like entries in a cache), cleaning up those files during the step will produce less information to save for that step.

Of course, you want to create checkpoints at certain stages: because layers are cached, if you can find some steps that don't change often, creating a layer means that you can reuse that state in future builds.  So don't just lump everything together either.  Build the image in stages: the most stable stuff comes first, the least stable stuff comes last.

PR #852, opened 2022-08-04
test: Fix tests to adapt bib.ietf.org

kesara:


tests/input/draft-template-old.xml
```diff
@@ -0,0 +1,583 @@
...
+     An alternate method (rfc include) is described in the references. -->
+
+<!ENTITY RFC2119 SYSTEM "http://xml2rfc.ietf.org/public/rfc/bibxml/reference.RFC.2119.xml">
+<!ENTITY RFC2629 SYSTEM "http://xml2rfc.ietf.org/public/rfc/bibxml/reference.RFC.2629.xml">
+<!ENTITY RFC3552 SYSTEM "http://xml2rfc.ietf.org/public/rfc/bibxml/reference.RFC.3552.xml">
+<!ENTITY I-D.narten-iana-considerations-rfc2434bis SYSTEM "http://xml2rfc.ietf.org/public/rfc/bibxml3/reference.I-D.narten-iana-considerations-rfc2434bis.xml">
+]>
+
+<!-- This template is for creating an Internet Draft using xml2rfc,
+     which is available here: http://xml.resource.org. -->
```
rjsparks, line 14: Even if this is an "old" artifact, we may want to clean up misleading stuff like that? I see also that this uses postal, yet silences warnings about it, so it's not really "legacy" input...

tests/input/draft-template-old.xml
```diff
@@ -0,0 +1,583 @@
...
+      <t>This becomes an Appendix.</t>
+    </section>
+
+    <section anchor="contributors" title="Contributors" numbered="no">
+      <t>This becomes an unnumbered section</t>
+    </section>
+
+    <!-- Change Log
+
+v00 2006-03-15  EBD   Initial version
```
rjsparks, line 568: I doubt this has been maintained?

tests/input/draft-template.xml
```diff
@@ -1,19 +1,18 @@
...
-<!ENTITY I-D.narten-iana-considerations-rfc2434bis SYSTEM "http://xml2rfc.ietf.org/public/rfc/bibxml3/reference.I-D.narten-iana-considerations-rfc2434bis.xml">
+<!DOCTYPE rfc [
+  <!ENTITY nbsp    "&#160;">
+  <!ENTITY zwsp   "&#8203;">
+  <!ENTITY nbhy   "&#8209;">
+  <!ENTITY wj     "&#8288;">
+  <!-- One method to get references from the online citation libraries.
+       There has to be one entity for each item to be referenced.
+       An alternate method (rfc include) is described in the references. -->
+  <!ENTITY RFC2119 SYSTEM "http://xml2rfc.ietf.org/public/rfc/bibxml/reference.RFC.2119.xml">
```
rjsparks, line 10: switch these over now?

rjsparks (approved):
I think this can go in as is, but see inline for possible future improvements.

tests/input/draft-template-old.xml
kesara: I have removed the misleading comment.

tests/input/draft-template-old.xml
kesara: I don't think so, I have removed that too.

tests/input/draft-template-old.xml
kesara: I think that's okay to silence these warnings. Otherwise, it's going to be another warning in the test logs.

tests/input/draft-template.xml
kesara: I'll update the reference URLs to `bib.ietf.org`.
And I'll keep the `draft-template-old.xml` URLs as it is.

kesara: I have updated the URL except for `draft-template-old.xml`.

tests/input/draft-template-old.xml
rjsparks: well, you missed my point on postal, but it's not important enough to continue to hold this up. My earlier approval stands.

PR #855, opened 2022-08-10
build: Optimize docker images

kesara:
As per comments in https://github.com/ietf-tools/xml2rfc/pull/849#issuecomment-1210075932, attempt to optimize the docker images generated by xml2rfc.

### DIFF
--- a/docker/base.Dockerfile
+++ b/docker/base.Dockerfile
@@ -2,7 +2,13 @@ FROM ubuntu:jammy
 LABEL maintainer="IETF Tools Team <tools-discuss@ietf.org>"
 
 ENV DEBIAN_FRONTEND noninteractive
+ENV LANG=en_US.UTF-8
+
 WORKDIR /root
+
+# .bashrc configuration
+RUN echo 'xml2rfc --version' >> ~/.bashrc && \
+    echo 'if [ -d ~/xml2rfc ]; then cd ~/xml2rfc; fi' >> ~/.bashrc
 
 # Install dependencies
 RUN apt-get update --fix-missing && \
@@ -22,7 +28,10 @@ RUN apt-get update --fix-missing && \
         python3.10-dev \
         python3-pip \
         python3.10-distutils && \
-    locale-gen en_US.UTF-8
+    locale-gen en_US.UTF-8 && \
+    rm -rf /var/lib/apt/lists/* /var/log/dpkg.log && \
+    apt-get autoremove -y && \
+    apt-get clean -y
 
 # Install required fonts
 RUN mkdir -p ~/.fonts/opentype && \
@@ -50,12 +59,6 @@ COPY xml2rfc ./xml2rfc
 
 # Build xml2rfc & finalize
 RUN make install && \
-    apt-get remove --purge -y software-properties-common wget unzip && \
-    apt-get autoclean && \
-    apt-get clean && \
     pip3 uninstall -y decorator dict2xml pypdf2 && \
     rm setup.py Makefile configtest.py requirements.txt && \
-    rm -r xml2rfc build dist && \
-    echo 'export LANG=en_US.UTF-8' >> ~/.bashrc && \
-    echo 'xml2rfc --version' >> ~/.bashrc && \
-    echo 'if [ -d ~/xml2rfc ]; then cd ~/xml2rfc; fi' >> ~/.bashrc
+    rm -r xml2rfc build dist
--- a/docker/base.Dockerfile
+++ b/docker/base.Dockerfile
@@ -2,75 +2,60 @@ FROM ubuntu:jammy
 LABEL maintainer="IETF Tools Team <tools-discuss@ietf.org>"
 
 ENV DEBIAN_FRONTEND noninteractive
-
-RUN apt-get update --fix-missing
-
-# Install dependencies
-RUN apt-get install -y --fix-missing \
-    libpango-1.0-0 \
-    libssl-dev \
-    fontconfig \
-    libharfbuzz0b \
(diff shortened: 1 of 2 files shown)


kesara:
@martinthomson I can't add you as a reviewer because of the project restrictions to stop spam, but appreciate your comments on this PR.

docker/base.Dockerfile
```diff
@@ -2,75 +2,60 @@ FROM ubuntu:jammy
...
+RUN make install && \
+    apt-get remove --purge -y software-properties-common wget unzip && \
+    apt-get autoclean && \
+    apt-get clean && \
+    pip3 uninstall -y decorator dict2xml pypdf2 && \
+    rm setup.py Makefile configtest.py requirements.txt && \
+    rm -r xml2rfc build dist && \
+    echo 'export LANG=en_US.UTF-8' >> ~/.bashrc && \
+    echo 'xml2rfc --version' >> ~/.bashrc && \
+    echo 'if [ -d ~/xml2rfc ]; then cd ~/xml2rfc; fi' >> ~/.bashrc
```
NGPixel, line 61: Should be a separate RUN statement, as it's not related to what's above.

docker/dev.Dockerfile
```diff
@@ -2,46 +2,32 @@ FROM ghcr.io/ietf-tools/xml2rfc-base:latest
...
-
-# cleanup
-RUN apt-get remove --purge -y software-properties-common
-RUN apt-get autoclean
-RUN apt-get clean
-
-# bash config
-RUN echo 'xml2rfc --version --verbose' >> ~/.bashrc
+    "pypdf2>=2.6.0" && \
+    echo 'xml2rfc --version --verbose' >> ~/.bashrc
```
NGPixel, line 31: Should be a separate RUN statement, like above.

NGPixel (changes requested):

docker/base.Dockerfile
```diff
@@ -2,75 +2,60 @@ FROM ubuntu:jammy
...
+        libxml2-utils \
+        groff \
+        wget \
+        unzip \
+        locales \
+        python3.10 \
+        python3.10-dev \
+        python3-pip \
+        python3.10-distutils && \
+    locale-gen en_US.UTF-8
```
martinthomson, line 25: To remove cached and temporary files, add:
```
 && rm -rf /var/lib/apt/lists/* /var/log/dpkg.log \
 && apt-get autoremove -y && apt-get clean -y
```

docker/base.Dockerfile
martinthomson: Yeah, and you could move the last three lines further up as well (they are not dependent on anything that happens at this later stage.  (Before the COPY line even.)

martinthomson (commented):
A few tips, but this should be a lot faster and leaner as-is.

docker/base.Dockerfile
```diff
@@ -2,75 +2,60 @@ FROM ubuntu:jammy
...
-RUN echo 'if [ -d ~/xml2rfc ]; then cd ~/xml2rfc; fi' >> ~/.bashrc
+# Build xml2rfc & finalize
+RUN make install && \
+    apt-get remove --purge -y software-properties-common wget unzip && \
+    apt-get autoclean && \
+    apt-get clean && \
+    pip3 uninstall -y decorator dict2xml pypdf2 && \
+    rm setup.py Makefile configtest.py requirements.txt && \
+    rm -r xml2rfc build dist && \
+    echo 'export LANG=en_US.UTF-8' >> ~/.bashrc && \
```
martinthomson, line 59: This might be an ENV instruction.

docker/base.Dockerfile
```diff
@@ -2,75 +2,60 @@ FROM ubuntu:jammy
...
-RUN rm setup.py Makefile configtest.py requirements.txt
-RUN rm -r xml2rfc build dist
-
-# bash config
-RUN echo 'export LANG=en_US.UTF-8' >> ~/.bashrc
-RUN echo 'xml2rfc --version' >> ~/.bashrc
-RUN echo 'if [ -d ~/xml2rfc ]; then cd ~/xml2rfc; fi' >> ~/.bashrc
+# Build xml2rfc & finalize
+RUN make install && \
+    apt-get remove --purge -y software-properties-common wget unzip && \
```
martinthomson, line 53: Removing things doesn't really help save on space.  It increases image size and build time, contrary to what you might expect.  Besides, these seem like useful tools to have :)

The way to build a software package from source without inflating image size is a little counter-intuitive.  Do that by creating two images - a big one and a small one - and copying the build product from the big image to the small one.  See https://github.com/mozilla/neqo/blob/main/qns/Dockerfile

martinthomson (commented):
Now for the fun comments...

It really depends on how far down the rabbit hole you want to get.

docker/base.Dockerfile
kesara: I've moved these statements to the top.

docker/dev.Dockerfile
kesara: This line wasn't necessary, so I remove it.

docker/base.Dockerfile
kesara: I've added these commands to both base and dev Dockerfiles.

docker/base.Dockerfile
kesara: I don't want to move to that system yet, so I've removed the `apt-get remove` command.

martinthomson: Yeah, it's a fairly big lift and probably overkill for what you need here.

NGPixel (approved):

PR #867, opened 2022-08-18
Add classes to xref

martinthomson:
For #663.  I think that this might be everything aside from adding a class for normative and non-normative; that part is tricky to implement.

The classes I have managed to implement work as follows:

* Most links here will get the "xref" class. (as before)
* Relative links to sections of citations get "relref", but not "xref". (as before)
* Links to citations/references (like "[RFC2119]") get the "cite" class. (new)
* Links to sections (or appendices) of this document get the "internal" class. (new)
* Links with default text (like "Section 1.3") get the "auto" class. (new)
* Links with format="none" get the "plain" class. (new)

This should make styling much easier to do.

I noted a quirk: `<xref target="blah" format="none"/>` is permitted by xml2rfc, even though it does nothing.  That might be worth a warning.  I noticed this because, as implemented here, that is marked "auto", which is odd, because other format="none" links are marked "plain xref".  I'll open a bug.


xml2rfc/writers/html.py
```diff
@@ -2808,7 +2808,14 @@ def render_uri(self, h, x):
...
     # 
     #    [<a class="xref" href="#RFC1234">RFC1234</a>]
-    # 
+    #
+    # Regarding HTML class assignment:
+    #   Most links here will get the "xref" class.
+    #   Relative links to sections of citations get "relref", but not "xref".
+    #   Links to citations/references (like "[RFC2119]") get the "cite" class.
+    #   Links to sections (or appendices) of this document get the "internal" class.
+    #   "sref" links with default text (like "Section 1.3") get the "auto" class.
```
cabo, line 2817: What is sref?

martinthomson: Sorry, I messed that up (it was a name I used for "internal" previously).

```suggestion
    #   Links with default text (like "Section 1.3") get the "auto" class.
```

kesara (approved):

PR #884, opened 2022-08-29
 fix: RTL unicode issue in PDF

### LINKED ISSUE #873: RTL breakage in PDF
### Describe the issue

Archived-At: <https://mailarchive.ietf.org/arch/msg/auth48archive/_8ZjoHcPh5HfYxRQb3P4zVbdXUg>

  https://www.rfc-editor.org/authors/rfc9290.txt
  https://www.rfc-editor.org/authors/rfc9290.pdf
  https://www.rfc-editor.org/authors/rfc9290.html
  https://www.rfc-editor.org/authors/rfc9290.xml

This is likely to lead to TI state.

### Code of Conduct

kesara:
Fixes #873

### DIFF
--- a/xml2rfc/writers/pdf.py
+++ b/xml2rfc/writers/pdf.py
@@ -5,6 +5,7 @@ from __future__ import unicode_literals, print_function, division
 import io
 import logging
 import os
+import re
 
 import warnings
 
@@ -77,6 +78,8 @@ class PdfWriter(BaseV3Writer):
         htmlwriter = HtmlWriter(self.xmlrfc, quiet=True, options=self.options, date=self.date)
         html = htmlwriter.html()
 
+        html = self.flatten_unicode_spans(html)
+
         writer = weasyprint.HTML(string=html, base_url="")
 
         cssin  = self.options.css or os.path.join(os.path.dirname(os.path.dirname(__file__)), 'data', 'xml2rfc.css')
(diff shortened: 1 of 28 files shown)


xml2rfc/writers/pdf.py
```diff
@@ -61,6 +62,14 @@ def __init__(self, xmlrfc, quiet=None, options=default_options, date=None):
         else:
             ftlogger.setLevel(logging.ERROR)
 
+    def flatten_unicode_spans(self, html):
+        # This is a fix for bug in WeasyPrint that doesn't handle RTL unicode
+        # content correctly.
+        # See #873 & Kozea/WeasyPrint#1711
+        return re.sub(r'<span class="unicode">(?P<unicode_content>.*)</span>',
```
cabo, line 69: Shouldn't this be a `.*?` so it doesn't bleed over into additional spans?
Maybe add a test case with `<u>a</u>b<u>c</u>` as well.

kesara: Yes, you are correct. It's greedy by default. I'll fix that and add an unit test.

kesara: I've updated the commit.

cabo: LGTM now.

rjsparks (approved):

Issue #833, opened 2022-07-14
Issues with colspan (.txt output) and shading (.html, .pdf outputs) when trying to convert a figure to a table


lbartholomew-rpc:
Issues with colspan (.txt output) and shading (.html, .pdf outputs) when trying to convert a figure to a table (Table 1 in the "-with-Table1" files listed below).

The "colspan" entries had several undesirable carriage returns each in the .txt output (the "BIFT indices for Packets with SI=0" and "BIFT indices for BitString/Packet with SI=1" rows).

Please see

https://www.rfc-editor.org/v3test/draft-ietf-bier-te-arch-13-with-Table1.xml
https://www.rfc-editor.org/v3test/draft-ietf-bier-te-arch-13-with-Table1.txt
https://www.rfc-editor.org/v3test/draft-ietf-bier-te-arch-13-with-Table1.html
https://www.rfc-editor.org/v3test/draft-ietf-bier-te-arch-13-with-Table1.pdf

The shading in the 4th row in the .html and .pdf outputs was off, but no carriage-return issue for the colspan entries.

To see what the correct layout should be, please see Figure 4 in

https://www.rfc-editor.org/v3test/draft-ietf-bier-te-arch-13-xml2.xml
https://www.rfc-editor.org/v3test/draft-ietf-bier-te-arch-13-xml2.txt

alicerusso:
Seems to be another example of the colspan issue.  Attempting to use `<table>` to create Table 5 in this document: https://www.ietf.org/id/draft-irtf-nmrg-ibn-intent-classification-08.txt
In the text output, see the 5 right-most columns --`<th>` is being forced into one column even though it has colspan set greater than 1 (e.g., `<th colspan="6">Intent Scope</th>`). 

Files:
https://www.rfc-editor.org/v3test/colspan-table-example.xml
https://www.rfc-editor.org/v3test/colspan-table-example.txt
https://www.rfc-editor.org/v3test/colspan-table-example.html
https://www.rfc-editor.org/v3test/colspan-table-example.pdf

original:
```
   +---------+-------------+-----------------+-----+-----+-----+-----+
   |Intent   | Intent      | Intent          | DCN | DCN | ABS | L-C |
   |User     | Type        | Scope           | Res | Net |     |     |
   |         |             +-----------------+-----+-----+-----+-----+
   |         |             |C1|C2|C3|C4|C5|C6|C1|C2|C1|C2|C1|C2|C1|C2|
[...]

PR #886, opened 2022-08-31
 fix: Fix table colspan issue in text format 

kesara:
Fix for table colspan issue in #833


rjsparks (approved):

Doondondon (commented):
// endpoint for all activity

PR #895, opened 2022-09-15
Turn the beyond-ascii check into a warning for RFC mode

cabo:
(See discussion in auth48archive and others.)


rjsparks:
So far this will make it so that non-ascii utf-8 can appear in `t` without any warning being emitted. The warning/error changes would downshift the error to a warning for other places xml2rc was being strict about utf-8 for.

I'm not sure that's the intended result of the PR?

Unless we have some other tool to let the rfc-editor know they've got utf-8 to look at for reasonableness in the chain, having it pass it quietly seems a step too far.

cabo:
The change only kicks in when the RFC has a number.  Up to then, xml2rfc works like before.

rjsparks:
and sometimes big swaths of text get added after the RFC has a number.

cabo:
Yes, and having a tool that gives the RFC editor a histogram of character usage is probably quite useful.  SMOP.

cabo:
``` ruby
require 'unicode/name'
hist = Hash.new(0)
ARGF.read.each_char do |c|
  hist[c] += 1 unless c.ord < 128
end
cl = "*** Latin"
hist.keys.sort.partition {|c| c =~ /\A\p{Latin}\z/}.each do |l|
  puts cl
  l.each  do |c|
    puts "#{c}: #{"%4d" % hist[c]} #{Unicode::Name.correct(c)}"
  end
  cl = "*** Non-Latin"
end
```

```
$ unicode-histo rfc9277.xml
*** Latin
ß:    1 LATIN SMALL LETTER SHARP S
ä:    2 LATIN SMALL LETTER A WITH DIAERESIS
ö:    2 LATIN SMALL LETTER O WITH DIAERESIS
*** Non-Latin
 :   52 NO-BREAK SPACE
≤:    2 LESS-THAN OR EQUAL TO
```

sginoza:
I need to think about this a bit more but 2 things come to mind: 
a) slightly concerned that this isn't getting wider review in that this has been normal handling since the transition to v3 (so almost 3 years)
b) if we're making the switch to allow non-ascii everywhere, flagging if and where these characters appear would be helpful.  I want to talk to the team about this to see what is helpful for all, but my first thought is it would be helpful to have some of the surrounding text or line numbers so we know where the characters appear (maybe some instances are ok and some aren't).   We have a script we use now, but having something built into xml2rfc might be nice to have one less separate check to run.

xml2rfc/writers/preptool.py
```diff
@@ -510,16 +510,22 @@ def check_ascii_text(self, e, p):
                     if not c.get('ascii') and not c.tag in bare_unicode_tags and not is_script(c.text, 'Latin'):
                         self.err(c, 'Found non-ascii content without matching ascii attribute in <%s>: %s' % (c.tag, show))
                 else:
-                    self.err(c, 'Found non-ascii characters outside of elements that can have non-ascii content, in <%s>: %s' % (c.tag, show))
-                    c.text = downcode(c.text)
+                    if self.draft:
+                        self.err(c, 'Found non-ascii characters outside of elements that can have non-ascii content, in <%s>: %s' % (c.tag, show))
+                        c.text = downcode(c.text)
+                    else:
+                        self.warn(c, 'Found non-ascii characters outside of elements that can have non-ascii content, in <%s>: %s' % (c.tag, show))
```
kesara, line 517: `xml2rfc` will not hit this warning for document with RFC number that has Unicode content on `<t>` because `enable_rfc_mode_rfc7997_treatment()` adds `<t>` to `bare_unicode_tags` and `unicode_content_tags`.

xml2rfc/writers/preptool.py
```diff
@@ -510,16 +510,22 @@ def check_ascii_text(self, e, p):
...
                     if not p.get('ascii') and not p.tag in ['artwork', 'refcontent', 'u']:
                         self.err(p, 'Found non-unicode content without matching ascii attribute in <%s>: %s' % (p.tag, show))
                 else:
-                    self.err(p, 'Found non-ascii characters outside of elements that can have non-ascii content, in <%s>: %s' % (p.tag, show))
-                    c.tail = downcode(c.tail)
+                    if self.draft:
+                        self.err(p, 'Found non-ascii characters outside of elements that can have non-ascii content, in <%s>: %s' % (p.tag, show))
+                        c.tail = downcode(c.tail)
+                    else:
+                        self.warn(p, 'Found non-ascii characters outside of elements that can have non-ascii content, in <%s>: %s' % (p.tag, show))
```
kesara, line 528: Same issue as https://github.com/ietf-tools/xml2rfc/pull/895/files#r972501724

xml2rfc/writers/preptool.py
cabo: Correct.  The change directly to bare_unicode_tags and unicode_content_tags probably obviates making the change to preptool.py.  I was hoping that the prep tool change was sufficient, also because of the convenient warning, but xml2rfc then simply downcodes the text, which is not useful.

kesara:
@rjsparks, @cabo: I have commited a new change to allow bare Unicode in the `<t>` element.
This also introduces new flag `--warn-bare-unicode` when set, `xml2rfc` warns about bare Unicode in the `<t>` element.
By default, this is set to False.

Have a look at the following test results (in a different branch, I haven't merged the results to this commit yet): https://github.com/kesara/xml2rfc/commit/3998d8e1b142ca76d706b80ad6b7afe343fe8425#diff-c2d2944d758bb3948e7924d76edf0e05292cda738d52c51636723fbbd444cfa9

Are those test results acceptable?

xml2rfc/writers/preptool.py
```diff
@@ -507,25 +509,25 @@ def check_ascii_text(self, e, p):
             if c.text and not isascii(c.text):
                 show = c.text.encode('ascii', errors='replace')
                 if c.tag in unicode_content_tags:
-                    if not c.get('ascii') and not c.tag in bare_unicode_tags and not is_script(c.text, 'Latin'):
+                    if c.tag in bare_unicode_tags_with_notice:
+                        if self.options.warn_bare_unicode:
+                            self.warn(c, 'Found non-ascii characters outside of elements that can have non-ascii content, in <%s>: %s' % (c.tag, show))
```
cabo, line 514: The statement "can have" is unfounded.  But this is a warning, so maybe that is not a problem.

xml2rfc/run.py
```diff
@@ -181,6 +181,8 @@ def main():
                             help='show option values and from where they come')
     help_options.add_argument('-V', '--version', action='store_true', 
                             help='display the version number and exit')
+    help_options.add_argument('--warn-bare-unicode', action='store_true', default=False,
```
cabo, line 184: My original PR was liberating the RFC mode only.  The equivalent here would be to keep the warning in I-D mode (default=True), and disable it in RFC mode (default=False).  Of course, I'm happier with the code as it now is.

rjsparks: We had a long discussion during an rpat call that convinced me that xml2rfc itself shouldn't be gating the entry of utf-8 into `<t>` for Internet-Drafts. I'm planning to give the IESG a heads up that it can happen, and we need to train authors and chairs to be aware of what the RFC Editor will ultimately allow. The RPC is building clarified language around that, and it will end up on authors.ietf.org.

xml2rfc/writers/preptool.py
rjsparks: How about: "Found non-ascii characters in a tag that should be inspected to ensure they are intentional"

cabo: Sounds good.  (We have this weird definition of ASCII that includes a few beyond-ASCII characters such as NBSP, but that level of detail goes beyond the warning message.)

cabo: Oh, and it should be "element", not "tag"...
(The content is in the element, but not in the tag, which only has the attributes.)

kesara: I have updated the warning to:
```
Found non-ascii characters in an element that should be inspected to ensure they are intentional, in <%s>: %s
```
Example:
```
Warning: Found non-ascii characters in an element that should be inspected to ensure they are intentional, in <t>: b'The example shows how the Hebrew-language string ????'
```

cabo:
Here is another example why this needs to be fixed:

https://github.com/ietf-wg-jsonpath/draft-ietf-jsonpath-base/issues/274

xml2rfc/util/unicode.py
```diff
@@ -34,6 +34,7 @@
     'street',
     'title',
     'u',
+    't',
```
martinthomson, line 37: Please add, both here and below: dd, dt, li, blockquote, and any other block-level elements I missed.  Then add, both here and below, the "inline" elements: cref (?), em, eref (?), iref (?), relref, strong, sub, sup, tt, and xref.  (I didn't think much about the cross reference stuff, but I think they can contain text.)

I have immediate use for many of these.  I can wrap in `<t>` for some of the block-level elements, but not the inline-level ones.

martinthomson: Note that with a list this long, you might consider how the latter list (bare_unicode_tags_with_notice) can be merged into the former (unicode_content_tags).

reschke:
> The change only kicks in when the RFC has a number. Up to then, xml2rfc works like before.

That sounds counter-intuitive to me. If it's an error until AUTH48, how will it be used by draft authors?

cabo:
> > The change only kicks in when the RFC has a number. Up to then, xml2rfc works like before.
> 
> That sounds counter-intuitive to me. If it's an error until AUTH48, how will it be used by draft authors?

My original idea was to minimize the appearance of a change: use the workaround until RFC editing stage, and use the RFC editor's authority to decide what actually should be done.
However, that approach is no longer part of the proposal here.

xml2rfc/util/unicode.py
rjsparks: I agree with the goal of the request. But I don't think we can do that without re-engaging the RSAB for a revised interpretation, or get a doc from RSWG through the RSAB that is explicit.

duerst: Unfortunately, I have to agree with @rjsparks here. RFC 7997 allows non-ASCII in examples, but not e.g. in Mathematical formulæ or other cases in general text. Of course I agree with @martinthomson that it would be highly desirable to use non-ASCII for these cases, but I think that requires a change in policy.

duerst:
@cabo wrote:

> ```
> $ unicode-histo rfc9277.xml
> *** Latin
> ß:    1 LATIN SMALL LETTER SHARP S
> ä:    2 LATIN SMALL LETTER A WITH DIAERESIS
> ö:    2 LATIN SMALL LETTER O WITH DIAERESIS
> *** Non-Latin
>  :   52 NO-BREAK SPACE
> ≤:    2 LESS-THAN OR EQUAL TO
> ```

I think separating Latin and non-Latin is a good first step, but for the RPC, some other statistics may also be helpful:
* Counts per script (e.g. 4 HEBREW), not necessarily per character.
* Instances of non-normalized or confusable characters (implementation of this might take some effort).
* Instances of spaces (except for U+0020, the ASCII SPACE), marks,...

cabo:
On 21. Dec 2022, at 21:17, Robert Sparks ***@***.***> wrote:
> 
> 
> @rjsparks commented on this pull request.
> 
> In xml2rfc/util/unicode.py:
> 
> > @@ -34,6 +34,7 @@
>      'street',
>      'title',
>      'u',
> +    't',
> 
> I agree with the goal of the request. But I don't think we can do that without re-engaging the RSAB for a revised interpretation, or get a doc from RSWG through the RSAB that is explicit.

I’m repeating myself here: This fixes a bug in xml2rfc.

There is no policy change.
(This is unfortunate, but we have to make these steps one after another.)

Whether support for implementing the actual policy is added to xml2rfc or to some linter tool (or some diagnostic aid like the unicode-histo sketch I put in above) is orthogonal to fixing this bug.

Grüße, Carsten

rjsparks:
> I’m repeating myself here: This fixes a bug in xml2rfc.

I don't disagree with you, but it's clear that the nascent RSWG/RSAB engine feels it has a role.  I misremembered the status of this PR during the RSAB meeting at 116 (claiming it was already merged), and there were several people who were very upset about it.

So, yes, where Martin points is the right place to go. But I think trying to go there without nods from that new engine will make landing it take _longer_.

tests/valid/draft-template-old.html
```diff
@@ -910,7 +910,7 @@ <h1 id="rfc.references.2">
 <tr>
 <td class="reference"><b id="I-D.narten-iana-considerations-rfc2434bis">[I-D.narten-iana-considerations-rfc2434bis]</b></td>
 <td class="top">
-<a>Narten, T.</a> and <a>H. Alvestrand</a>, "<a href="https://tools.ietf.org/html/draft-narten-iana-considerations-rfc2434bis-09">Guidelines for Writing an IANA Considerations Section in RFCs</a>", Internet-Draft draft-narten-iana-considerations-rfc2434bis-09, March 2008.</td>
+<a title="Google">Alvestrand, H. T.</a> and <a title="IBM">T. Narten</a>, "<a href="https://tools.ietf.org/html/draft-narten-iana-considerations-rfc2434bis-09">Guidelines for Writing an IANA Considerations Section in RFCs</a>", Internet-Draft draft-narten-iana-considerations-rfc2434bis-09, March 2008.</td>
```
rjsparks, line 913: These are terrible, and we will have to revert them when the bibxml is repaired, but I don't unless we take the time to pull depending on an external bib service from the tests now, it's unavoidable.

reschke: The issue is reversed names, or adding the org...?

rjsparks: All issues for bibxml not xml2rfc - initials handling, reordering authors (see https://github.com/ietf-tools/datatracker/issues/4973), `<format` tags (later down).

Issues _for_ xml2rfc (will track elsewhere) - not depending on the bibxml service for testing.

kesara: This has been fixed with #959.
This PR has been rebased and re-generated valid tests with https://github.com/ietf-tools/xml2rfc/pull/895/commits/3eb346823d33dfd80fc7aed1211406513d79d2c1

rjsparks (approved):

PR #909, opened 2022-10-10
fix: Use noto fonts for non-latin unicode monospaced characters

### LINKED ISSUE #906: PDF font leak when non latin characters in monospace elements
### Describe the issue

When non-latin [1] characters are present in monospace outputs like `tt, code, pre, code`, **xml2rfc** uses system fonts rather than Noto fonts in PDF output.
This is can cause various font leakages in PDF output [2].

I think the fix here is to use **Noto** fonts for other scripts.

References:
[1] Characters that can not be represented by Roboto Mono font. Roboto supports Latin, Greek (partial) and Cyrillic scripts. See https://en.wikipedia.org/wiki/Roboto#Language_support
[2] #903 

### Code of Conduct

kesara:
Use Noto fonts when Roboto Mono font can not display monospaced Unicode characters.

Fixes #906

### DIFF
--- a/xml2rfc/writers/pdf.py
+++ b/xml2rfc/writers/pdf.py
@@ -88,8 +88,7 @@ class PdfWriter(BaseV3Writer):
 
         # fonts and page info
         fonts = self.get_serif_fonts()
-        # Check that we also have mono fonts
-        self.get_mono_fonts()
+        mono_fonts = self.get_mono_fonts()
         page_info = {
             'top-left': self.page_top_left(),
             'top-center': self.full_page_top_center(),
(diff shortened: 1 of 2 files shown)


rjsparks (approved):

xml2rfc/writers/pdf.py
```diff
@@ -166,6 +174,9 @@ def flatten_unicode_spans(self, html):
     font-family: {fonts};
     width: 100%;
   }}
+  tt, code, pre, code {{
```
cabo, line 161: Reason for having `code` in here twice?

kesara: Good spotting! It's redundant.
Same in https://github.com/ietf-tools/xml2rfc/blob/main/xml2rfc/data/xml2rfc.css#L790

kesara: Fixed with #916.

PR #913, opened 2022-10-12
Drop dependency on kitchen

Flowdalic:
Kitchen is unmainted and should no longer be used, hence we this drops the dependency on kitchen.

PR #914, opened 2022-10-12
Use wcwidth to determine the monospace textual length of a string

Flowdalic:
This is supplementary to https://github.com/ietf-tools/xml2rfc/pull/913, which dropped the dependency on kitchen. Here we add a dependency on wcwidth, which is an maintained library to determine the monospace textual length of a string.


kesara (approved):

kesara:
@Flowdalic Thanks for the contributions. I'll merge this PR first because #913 needs this change.

PR #920, opened 2022-10-19
test: Fix PDF tests

kesara:


rjsparks (approved):

dkg:
I'm a bit confused by this change, as there is no pointer to any reported problem, either in the git commit messages or in the PR's description.  It seems like it is related to https://github.com/Kozea/WeasyPrint/issues/1752 which has already been fixed by weasyprint upstream.

dkg:
In general, if a new version of a dependency breaks xml2rfc, it is good practice to clearly report that failure to the dependency directly.  @kitterma in this case did the report (thanks!) but it'd be good if xml2rfc-specific workarounds could explicitly note the upstream bug report so that it's easier to figure out how to undo the changes when the dependency is fixed.

kitterma:
I think it's more than a little misleading to describe pinning a dependency to a specific version as a "Fix" in any case.  It definitely would have been useful if the issue had been reported to them when it was found.  This time, it turned out to be purely an upstream bug, not an intended change of behavior.  When Weasyprint 57.1 comes out you should be able remove the maximum version restriction.

For Debian, I've grabbed the fix and patched our package to address it until it is release: [Debian patch](https://salsa.debian.org/python-team/packages/weasyprint/-/commit/8a652a8b21ad72c1fdb59e1f55e6d1d93e96375d).

kesara:
Weasyprint 57.1 has been released with a fix.
I have opened a [PR](https://github.com/ietf-tools/xml2rfc/pull/932) to skip Weasyprint version 57.0 keeping 53.0 as the minimum version.

PR #931, opened 2022-11-04
Include OpenPGP certificates for signing the project in each release

dkg:
Over in
https://mailarchive.ietf.org/arch/msg/tools-discuss/vlZ7ToWeOHy16qOmvXjGfZmIHFQ @kesara wrote in an unsigned e-mail message:

> See https://github.com/ietf-tools/xml2rfc/releases/tag/v3.15.0 for
> release details.
>
> Note that, the PyPI release is signed by a new key that belongs to me.
> PGP: A50D 9882 197F 431A 42F8 ACBB 985F 7788 DC6C 78FB

For future planned signing key transitions, it's probably best practice to indicate it with a cryptographically strong announcement (this e-mail itself is not that).

This change includes the current OpenPGP certificate that we expect to be able to sign releases in the codebase itself, in XML2RFC-OPENPGP-KEYS.txt.

When you plan to change keys, add the new key to that same location (i.e., export two keys in a single ASCII-armored blob to this same file).  Make at least one release with the updated XML2RFC-OPENPGP-KEYS.txt, signed by the "old" key.  This provides a cryptographic trail for anyone who needs to confirm the handover.

Once you start making releases signed by the "new" key, you can update XML2RFC-OPENPGP-KEYS.txt to only include the current key.


kesara:
Thanks for this.
I have added this information to the release instructions: https://github.com/ietf-tools/xml2rfc/wiki/Release-Process#changing-signing-keys

kesara (approved):

PR #937, opened 2022-11-17
Font override pdf only

martinthomson:
This was starting to annoy me a little, so I spent five minutes on it.  I hope that this is acceptable.


xml2rfc/data/xml2rfc.css
```diff
@@ -16,6 +16,12 @@
 @import url('https://fonts.googleapis.com/css?family=Noto+Serif'); /* Serif (print) */
 @import url('https://fonts.googleapis.com/css?family=Roboto+Mono'); /* Monospace */
 
+:root {
+  --font-sans: 'Noto Sans', Arial, Helvetica, sans-serif;
+  --font-serif: 'Noto Serif', 'Times', 'Times New Roman', serif;
```
kesara, line 21: Is it intentional to introduce two new fallback fonts (Times & Times New Roman) here?

martinthomson: I'm doing that to be consistent with the fallbacks you have for the sans-serif variant.  This is a common fallback pattern.

PR #958, opened 2023-01-12
feat: Add editorial stream

### LINKED ISSUE #896: add submissionType for Editorial stream and SoTM in output
### Description

Please add a new submissionType value for the Editorial stream (created by RFC 9280).

For "Status of This Memo" content when rendering an RFC, please see item 25 ("Editorial Stream Informational") on this page:  https://www.rfc-editor.org/materials/status-memos.txt
which is the outcome of [S6.3 of RFC 9280](https://www.rfc-editor.org/rfc/rfc9280.html#section-6.3).

### Code of Conduct

kesara:
Add editorial stream as per #896 and [Section 6.3 of RFC 9280](https://www.rfc-editor.org/rfc/rfc9280.html#section-6.3).

Fixes #896

### DIFF
--- a/test.py
+++ b/test.py
@@ -8,6 +8,7 @@ import unittest
 import xml2rfc
 import xml2rfc.utils
 
+from xml2rfc.boilerplate_rfc_7841 import boilerplate_rfc_status_of_memo
 from xml2rfc.walkpdf import xmldoc
 from xml2rfc.writers.base import default_options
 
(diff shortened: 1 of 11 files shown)


test.py
```diff
@@ -582,5 +582,55 @@ def test_set_font_family(self):
...
+    category="info">
+</rfc>''')
+        self.writer.options.rfc = True
+        self.writer.root = rfc
+
+        self.writer.boilerplate_insert_status_of_memo(rfc, rfc)
+        self.assertEqual(len(self.writer.errors), 0)
+        self.assertEqual(len(rfc.xpath('./section/name[text()="Status of This Memo"]')), 1)
+        paras = rfc.xpath('./section/t')
+        self.assertIn('This document is not an Internet Standards Track specification; it is', paras[0].text)
```
rjsparks, line 624: * good that we have this test - I'm not finding where we're testing that for the other boilerplate types?
* I worry that this is a little fragile in that it could pass if chunks of text were somehow reordered? Could this be done with a single assert per paragraph instead?

rjsparks (approved):
Minor comments in the code - nothing blocking.
(I agree that #895 should come in first, and the changes we are exploring there would be reflected here).

test.py
kesara: Is this better? https://github.com/ietf-tools/xml2rfc/pull/958/commits/e757754c4816270a212d87e8bb252b30a3121226

Issue #961, opened 2023-01-24
one extraneous space in text output of figure name


alicerusso:
### Describe the issue

input:
```
        <figure anchor="Reply_message">
          <name>Reply Message Consisting of a Fixed Header, a Request Block TLV, Report     
Block TLV(s), a Name TLV, a Reply Block TLV, and Reply Sub-Block TLV(s)</name>
```

Text output:
```
      Figure 8: Reply Message Consisting of a Fixed Header, a Request
       Block TLV, Report Block TLV(s), a Name TLV, a Reply Block TLV,
                        and Reply Sub- Block TLV(s)
```

Desired output (no space after the hyphen):
```
      Figure 8: Reply Message Consisting of a Fixed Header, a Request
       Block TLV, Report Block TLV(s), a Name TLV, a Reply Block TLV,
                        and Reply Sub-Block TLV(s)
```

files:
https://www.rfc-editor.org/v3test/test9344.xml
https://www.rfc-editor.org/v3test/test9344.txt
(PDF and HTML are not affected.)

### Code of Conduct

- [X] I agree to follow the [IETF's Code of Conduct](https://github.com/ietf-tools/.github/blob/main/CODE_OF_CONDUCT.md)

kesara:
I have two code change proposals for this:
1. #965 Avoid breaking lines by hyphens. - This affects all text elements. I prefer this but this may have undesirable changes. _See the text changes in the PR._
2. #966 Include the figure prefix before rending the name text. - This only affects `figure > name`. Note this change will also change how Figure names are wrapped in the past. _See the text changes in the PR._

Other work around is to use **non-breaking hyphen** (`&#8209;`). This doesn't need any code changes.

cabo:
The part that I don't understand is why the figure name (apparently) is formatted to a width and then reflowed for a different width.
NBHY indeed is a workaround until we figure that out.

PR #965, opened 2023-02-02
fix: Avoid breaking text on hyphens 

kesara:
Fixes https://github.com/ietf-tools/xml2rfc/issues/961

### DIFF
--- a/tests/valid/draft-template.pages.text
+++ b/tests/valid/draft-template.pages.text
@@ -24,8 +24,8 @@ Status of This Memo
 
    Internet-Drafts are working documents of the Internet Engineering
    Task Force (IETF).  Note that other groups may also distribute
-   working documents as Internet-Drafts.  The list of current Internet-
-   Drafts is at http://datatracker.ietf.org/drafts/current/.
+   working documents as Internet-Drafts.  The list of current
+   Internet-Drafts is at http://datatracker.ietf.org/drafts/current/.
 
    Internet-Drafts are draft documents valid for a maximum of six months
    and may be updated, replaced, or obsoleted by other documents at any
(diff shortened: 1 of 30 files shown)


cabo:
Wow, this change seems to have a wide-ranging effect.
Do we really want to stop breaking on hyphens everywhere?
Note that Unicode has a non-breaking hyphen character ‑, U+2011, for where this is actually desired.

cabo:
tests/valid/draft-template.pages.text
has a nice example for why this change is generally undesirable.

tests/valid/draft-template.pages.text
```diff
@@ -565,10 +565,10 @@ Internet-Draft              Abbreviated Title                 March 2009
    [I-D.narten-iana-considerations-rfc2434bis]
               Alvestrand, H. T. and T. Narten, "Guidelines for Writing
               an IANA Considerations Section in RFCs", Work in Progress,
-              Internet-Draft, draft-narten-iana-considerations-
-              rfc2434bis-09, March 26, 2008,
-              <https://www.ietf.org/archive/id/draft-narten-iana-
-              considerations-rfc2434bis-09.txt>.
+              Internet-Draft,
+              draft-narten-iana-considerations-rfc2434bis-09, March 26,
+              2008, <https://www.ietf.org/archive/id/draft-narten-iana-c
```
cabo, line 570: See https://github.com/ietf-tools/xml2rfc/pull/965/#issuecomment-1413237933

kesara: Yeah, this is ugly.

kesara:
> Wow, this change seems to have a wide-ranging effect. Do we really want to stop breaking on hyphens everywhere? Note that Unicode has a non-breaking hyphen character ‑, U+2011, for where this is actually desired.

@cabo This change was just a proposal for review. I preferred this because this affects the whole text wrapping. But as you pointed out there are undesirable effects on URLs, but code can be changed to treat URLs differently.

See https://github.com/ietf-tools/xml2rfc/issues/961#issuecomment-1413347377
We better to move this conversation to that issue until we figure out which way to go.

rjsparks:
convert this to draft?

kesara:
Closing this as this is not the desired solution.

PR #966, opened 2023-02-02
fix: Include figure prefix in name before rendering text

kesara:
Fixes #https://github.com/ietf-tools/xml2rfc/issues/961


cabo:
Is the same fix needed for all the other uses of `.tjoin()`?
If not, why not?

kesara:
> Is the same fix needed for all the other uses of `.tjoin()`? If not, why not?

`tjoin()` is used to render `reference`, `references -> name`, `section -> name`, `table -> name`.
I'll experiment and see if this fix is adversely affecting any of those.

cabo:
> I'll experiment and see if this fix is adversely affecting any of those.

I just thought that they may have the same bug that Alice reported (line-breaking hyphens and putting them back together with a superfluous space), which would need to be fixed in a similar way.

PR #993, opened 2023-05-28
Include page numbers in the TOC when generating PDF.

teythoon:
Previously, the table of contents only included the section numbers and a link into the document.  Neither is really helpful when navigating a printout.

Add page numbers to the table of contents in the generated PDF.

### DIFF
--- a/xml2rfc/writers/pdf.py
+++ b/xml2rfc/writers/pdf.py
@@ -215,6 +215,12 @@ page_css_template = """
       margin-top: 1em;
     }}
   }}
+  #toc nav ul li p a:nth-child(2)::after {{
+    content: target-counter(attr(href), page);
+    position: absolute;
+    right: 0px;
+    text-indent: 0px;
+  }}
 }}
 """
 


kesara:
@teythoon Thanks for the PR.
The page numbers are in blue but not clickable at the moment.
Is there a way to make them hyperlinks?


Screenshot 2023-06-02 at 7 32 56 AM

teythoon:
Turns out my inserted element inherited the `text-indent` property, and resetting that made the inserted element's bounding box better reflect the content.  And with that, the links are clickable in PDF.js and poppler-based viewers.  Please re-check!

kesara:
> Turns out my inserted element inherited the `text-indent` property, and resetting that made the inserted element's bounding box better reflect the content. And with that, the links are clickable in PDF.js and poppler-based viewers. Please re-check!

Awesome, that seemed to have fixed the issue with the links.
The [RFC Production Advisory Team (rpat)](https://datatracker.ietf.org/group/rpat/about/) will discuss about this change.

kesara:
This change is inline [RFC7995](https://www.rfc-editor.org/rfc/rfc7995.html#section-3.1.8).

PR #998, opened 2023-06-05
fix: Update for google-i18n-address 3.0.0

### LINKED ISSUE #997: Incompatible with google-i18n-address 3.0.0
### Describe the issue

In the last few days, version 3.0.0 of google-i18n-address was released. It seems that xml2rfc is not yet compatible with this. Need to fix the incompatibility or at least pin the dependency in the short term.

### Code of Conduct

jennifer-richards:
Fixes #997

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,6 +1,6 @@
 appdirs>=1.4.4
 configargparse>=1.2.3
-google-i18n-address>=2.3.2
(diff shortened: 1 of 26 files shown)


cabo:
This needs an expedited release.

rjsparks (approved):

PR #1004, opened 2023-06-14
fix: Remove old tools server from BibXML lookup locations

### LINKED ISSUE #1002: Remove old tool server URLs from bibxml lookup locations
### Description

URLs like `https://xml2rfc.ietf.org/public/rfc/` and `https://xml2rfc.tools.ietf.org/public/rfc/` now redirects to `https://bib.ietf.org`.

### Code of Conduct

kesara:
Fixes #1002

### DIFF
--- a/xml2rfc/parser.py
+++ b/xml2rfc/parser.py
@@ -55,11 +55,7 @@ class CachingResolver(lxml.etree.Resolver):
                  templates_path=base.default_options.template_dir, verbose=None, quiet=None,
                  no_network=None, network_locs= [
                      'https://bib.ietf.org/public/rfc/',
-                     'https://xml2rfc.ietf.org/public/rfc/',
-                     'https://xml2rfc.tools.ietf.org/public/rfc/',
                      'http://bib.ietf.org/public/rfc/',
-                     'http://xml2rfc.ietf.org/public/rfc/',
-                     'http://xml2rfc.tools.ietf.org/public/rfc/',
                  ],
                  rfc_number=None, options=base.default_options):
         self.quiet = quiet if quiet != None else options.quiet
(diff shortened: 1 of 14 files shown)


xml2rfc/parser.py
```diff
@@ -55,11 +55,7 @@ def __init__(self, cache_path=None, library_dirs=None, source=None,
                  templates_path=base.default_options.template_dir, verbose=None, quiet=None,
                  no_network=None, network_locs= [
                      'https://bib.ietf.org/public/rfc/',
-                     'https://xml2rfc.ietf.org/public/rfc/',
-                     'https://xml2rfc.tools.ietf.org/public/rfc/',
                      'http://bib.ietf.org/public/rfc/',
```
rjsparks, line 58: Why look here? If the object exists the https: variant will return it. If it doesn't we just get 404's twice.

xml2rfc/parser.py
```diff
@@ -481,11 +475,7 @@ def __init__(self, source, verbose=None, quiet=None, options=base.default_option
                  cache_path=None, templates_path=base.default_options.template_dir, library_dirs=None, add_xmlns=False,
                  no_network=None, network_locs=[
                      'https://bib.ietf.org/public/rfc/',
-                     'https://xml2rfc.ietf.org/public/rfc/',
-                     'https://xml2rfc.tools.ietf.org/public/rfc/',
                      'http://bib.ietf.org/public/rfc/',
```
rjsparks, line 478: same as above

rjsparks (changes requested):

xml2rfc/parser.py
kesara: Yeah, because `http://bib.ietf.org/public/rfc/` redirects to `https` it doesn't make sense to have it here. I'll remove that too.

rjsparks (approved):

PR #1005, opened 2023-06-16
fix: Add xml extension to BibXML requests

### LINKED ISSUE #1003: Add .xml extension when BibXML lookups with label.
### Description

For BibXML references without an extension like: 
```
<?rfc include='reference.I-D.ietf-httpbis-messaging'?>
```

add the `.xml` extension so BibXML lookup doesn't fail.

Ref: https://mailarchive.ietf.org/arch/msg/tools-discuss/QVrRuZ2HRq7pwO-ZenW9oSr3Q9A/

### Code of Conduct

kesara:
This change adds a `.xml` extension when the `.xml` extension is missing in BibXML reference requests.

Fixes #1003

### DIFF
--- a/xml2rfc/parser.py
+++ b/xml2rfc/parser.py
@@ -227,12 +227,10 @@ class CachingResolver(lxml.etree.Resolver):
                     result = os.path.join(self.source_dir, basename)
                     attempts.append(result)
         else:
-            if self.options and self.options.vocabulary == 'v3':
-                paths = [ request ]
-            elif not request.endswith('.xml'):
-                paths = [ request, request + '.xml' ]
+            if not request.endswith('.xml'):
+                paths = [ request + '.xml', request ]
             else:
-                paths = [ request ]                
+                paths = [ request ]
             if os.path.isabs(paths[0]):
                 # Absolute path, return as-is
                 for path in paths:
--- a/tests/input/draft-template.xml
+++ b/tests/input/draft-template.xml
@@ -553,6 +553,8 @@ main(int argc, char *argv[])
           <date year="1972" />
         </front>
       </reference>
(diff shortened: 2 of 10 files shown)


xml2rfc/parser.py
```diff
@@ -227,12 +227,10 @@ def getReferenceRequest(self, request, include=False, line_no=0):
                     result = os.path.join(self.source_dir, basename)
                     attempts.append(result)
         else:
-            if self.options and self.options.vocabulary == 'v3':
-                paths = [ request ]
-            elif not request.endswith('.xml'):
-                paths = [ request, request + '.xml' ]
+            if not request.endswith('.xml'):
```
rjsparks, line 230: On a quick read, I think this is likely to _break_ anyone that happened to be using files locally that didn't have the .xml extension - maybe there aren't any of those, but how would we know?. I think you want to put both request and request with .xml added into the list instead.

rjsparks: This is also (or was also) the pure v3 codepath - do we really want to start second-guessing what we're given as v3?

kesara: > On a quick read, I think this is likely to _break_ anyone that happened to be using files locally that didn't have the .xml extension - maybe there aren't any of those, but how would we know?. I think you want to put both request and request with .xml added into the list instead.

No, it's not going to break for those cases. Because next line adds both variants to the path with and without extension.

kesara: > This is also (or was also) the pure v3 codepath - do we really want to start second-guessing what we're given as v3?

Original code was not looking at the version attribute rather rely on what's set in command line options to figure out it was v3. I think this doesn't do any harm. Also #1006 adds a warning for use of PIs in v3.

rjsparks: (reticent) ok

rjsparks (approved):

PR #1011, opened 2023-06-22
fix: Reset seen_slugs cache on every prep run

jennifer-richards:
The `writers.preptool.slugify_name()` helper kept a set of `seen_slugs` as a module-level variable. As a result, uniqueness among slugs was enforced for the entire lifetime of a module import, including across `PrepToolWriter` instances and across multiple calls to `prep()`.

In xml2rfc itself, this appears to have been adding unnecessary "-2" suffixes on many HTML ids, at least in the tests. For datatracker's use of xml2rfc as a library, this was causing difficulties with long-lived celery workers by adding suffixes to prevent collisions between entirely separate drafts and eventually running out of unique slugs.

This makes the `slugify_name()` procedure a method on `PrepToolWriter` and resets the `seen_slugs` cache, now a property of the instance, on each invocation of `prep()`

### DIFF
--- a/tests/valid/rfc7911.html
+++ b/tests/valid/rfc7911.html
@@ -26,7 +26,7 @@
     intervaltree 3.1.0
     Jinja2 3.1.2
     lxml 4.9.2
-    platformdirs 3.6.0
+    platformdirs 3.7.0
     pycountry 22.3.5
     PyYAML 6.0
     requests 2.31.0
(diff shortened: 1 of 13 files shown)


tests/valid/rfc7911.html
```diff
@@ -1382,17 +1382,17 @@ <h2 id="name-introduction-2">
       convergence in a network by providing potential alternate or backup paths, 
       respectively.<a href="#section-1-3" class="pilcrow">¶</a></p>
 <section id="section-1.1">
-        <h3 id="name-specification-of-requirement">
-<a href="#section-1.1" class="section-number selfRef">1.1. </a><a href="#name-specification-of-requirement" class="section-name selfRef">Specification of Requirements</a>
+        <h3 id="name-specification-of-requiremen">
```
kesara, line 1385: `slugify_name` has removed a character from the ID.

kesara (approved):

PR #1021, opened 2023-08-03
fix: Break word when line overflows

### LINKED ISSUE #687: Text in HTML/PDF not wrapping properly
`type_defect`   |    by sginoza@amsl.com

___

There are some CBOR encodings in running text that is not getting wrapped in HTML and PDF.  The PDF is particularly bad because the text runs off the page, and there are no warnings that something odd is happening in the PDF.  The text is ok.  

https://www.rfc-editor.org/v3test/sandy9173.pdf
https://www.rfc-editor.org/v3test/sandy9173.html#appendix-A.1.1.3
https://www.rfc-editor.org/v3test/sandy9173.xml

The issue is sprinkled throughout the Appendix; see examples in sections A.1.1.3, A.1.3.2, and A.1.4. 

I'm not sure what the correct handling is supposed to be (e.g., warning vs rendering with line breaks).  Perhaps these should only be included as <sourcecode> or <artwork>.  

___
_Issue migrated from trac:687 at 2022-02-08 07:18:12 +0000_

kesara:
When a word is too big to fit in a line in HTML and PDF outputs, break the word.

Fixes #687

### DIFF
--- a/tests/valid/draft-miek-test.html
+++ b/tests/valid/draft-miek-test.html
@@ -87,6 +87,7 @@ body {
   font-family: var(--font-sans);
   line-height: 1.6;
   scroll-behavior: smooth;
+  overflow-wrap: break-word;
 }
 .ears {
   display: none;
@@ -655,7 +656,6 @@ hr.addr {
 /* pagination */
 @media print {
   body {
-
     width: 100%;
   }
   p {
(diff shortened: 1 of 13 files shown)


kesara:
> Code is fine, maybe change commit message to "when line overflows" instead of "when the overlap" for clarity?

I have updated the commit message.

PR #1033, opened 2023-09-21
feat(html): Add created datetime as a meta tag

### LINKED ISSUE #1032: HTML output: capture date of publication and date of file generation
### Description

Currently the HTML output captures the version of xml2rfc in the `<meta>` element, e.g., `<meta content="xml2rfc 3.17.4" name="generator">` and also in a comment:
```
<!-- Generator version information:
  xml2rfc 3.17.4
    Python 3.9.15
    ConfigArgParse 1.5.3
    google-i18n-address 3.0.0
    intervaltree 3.1.0
    Jinja2 3.1.2
    lxml 4.9.0
    platformdirs 3.8.0
    pycountry 22.3.5
    PyYAML 6.0
    requests 2.28.0
    setuptools 44.1.1
    six 1.16.0
    wcwidth 0.2.5
    weasyprint 56.1
-->
```
When the HTML output is created, it would be good to also have a timestamp of publication (that is, propagate the timestamp from the prepTime attribute found in the XML file) and a timestamp that captures when the HTML was rendered. If the HTML is then rerendered in the future, then these two timestamps will capture that info (that is, the publication timestamp will be older than the rendering timestamp). 
See the [RPAT meeting notes](https://notes.ietf.org/rpat-20230918#4-Review-of-HTML-files-created-by-xml2rfc-RFC-8650-and-beyond). 

### Code of Conduct

kesara:
This change adds the created datetime in the UTC time zone using ISO
datetime format for HTML output of an RFC.
Example:
```
  <meta content="2023-09-21T10:47:04.440964+00:00" name="created">
```

Fixes #1032


cabo:
As per the discussions in SEDATE, `+00:00` should be `Z` now...
(Unless you want to express the document was created in Reykjavik.)

kesara:
> As per the discussions in SEDATE, `+00:00` should be `Z` now... (Unless you want to express the document was created in Reykjavik.)

Python standard library is still not supporting `Z` [1] but I'll change the code so that `xml2rfc` adds `Z` instead of `+00:00`.

[1] https://github.com/python/cpython/issues/90772

kesara:
Updated created timestamp to show hours and minutes with `Z` instead of `+00:00`.
Example:
```
<meta content="2023-09-22T11:16Z" name="created">
```

PR #1047, opened 2023-10-11
fix: Respect newline attr on dl tag

### LINKED ISSUE #1045: xml2rfc v3.18.1:  Even with <dl newline="false">, get new line
### Describe the issue

xml2rfc v3.18.1:  Even with an explicit <dl newline="false">, get new line after the <dt> entry.  This was not an issue until 3.18.1.

### Code of Conduct

jennifer-richards:
Fixes #1045 

Bug was expecting `split(stripspace...)` to treat its argument the same way as `strip(stripspace)`. The latter strips any of the individual characters in `stripspace`. The former splits on _exactly_ the full string, not any of its characters. As a result, when calculating whether the first word of `<dd>` would fit alongside the `<dt>` text, the entire `<dd>` text was used. This was introduced in #1023.

A second bug caused the calculation actually to compare the length of the `<dd>` text to the full output width without actually accounting for the length of the `<dt>` text already on the line. This bug was present for a long time but would only manifest if the `<dt>` text plus the first word of `<dd>` should have overflowed the width, which would not be common.

This patch fixes both bugs and refactors the code to be more clear in its intention. The `newline` attribute should now match what's described at https://www.ietf.org/archive/id/draft-rswg-xml2rfcv3-implemented-02.html#section-3.20.3


jennifer-richards:
Bleah - the commit comment on a39b20a should probably say `<dd>` rather than `<dt>` (though it's sort of true either way).

rjsparks (approved):

PR #1050, opened 2023-10-26
fix: Deduplicate index entries

### LINKED ISSUE #988: Deduplicate index entries
### Describe the issue

Items can be entered into the index multiple times from the same paragraph.  xml2rfc happily creates multiple links to the same place.

It might be useful to remove such duplicates.  This might be done along with a future iteration of #862, where entries might be arranged more compactly at the same time.

### Code of Conduct

jennifer-richards:
This PR addresses two issues:
* Duplicate entries per term in the index.
* Duplicate terms in the index.

Fixes #988

### DIFF
--- a/tests/valid/docfile.html
+++ b/tests/valid/docfile.html
@@ -4,15 +4,15 @@
 <meta charset="utf-8">
 <meta content="Cherokee,Common,Greek,Latin" name="scripts">
 <meta content="initial-scale=1.0" name="viewport">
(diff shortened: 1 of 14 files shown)


kesara:
The root cause for having duplicate terms is some terms are stored with white spaces due to line breaks.
For example, there can be two unique entries because one term is `'Oblivious Gateway Resource'` and the other is `'Oblivious Gateway       Resource'` (due to line breaks).

kesara (approved):

rjsparks (approved):

PR #1055, opened 2023-11-16
fix: Follow HTML presentational hints in PDF

### LINKED ISSUE #1054: Numbering in <ol> not correct after <aside> and <artwork>
In the pdf output, numbering for an `<ol>` does not continue correctly after `<aside>` or `<artwork>`. The txt and html outputs are correct.

The "start" attribute of `<ol>` is used after the `<aside>` or `<artwork>` element (with the appropriate value of "start" to continue the numbering):

`<ol spacing="compact" start="4">`

A workaround for the misnumbering following `<artwork>` is to nest the `<artwork>` within `<li>` and not start a new `<ol>` with the start attribute. This doesn’t help after `<aside>` though, as `<aside>` is not allowed in `<li>` or `<ol>`.

For examples, see Sections 4.3.1, 4.3.2, 4.3.4, 5.3.1, 5.3.2, and 5.3.4 in the pdf output in these files: 

https://www.rfc-editor.org/v3test/rfc9496.pdf
https://www.rfc-editor.org/v3test/rfc9496.txt
https://www.rfc-editor.org/v3test/rfc9496.html
https://www.rfc-editor.org/v3test/rfc9496.xml

One example of incorrect numbering in pdf output (should be numbered 1, 2, 3, 4 rather than 1, 1, 2, 1):

```
   1.  Interpret the string as an unsigned integer s in little-endian
       representation.  If the length of the string is not 32 bytes or
       if the resulting value is >= p, decoding fails.

      |  Note: Unlike the field element decoding described in [RFC7748],
      |  the most significant bit is not masked, and non-canonical
      |  values are rejected.  The test vectors in Appendix A.2 exercise
      |  these edge cases.

   1.  If IS_NEGATIVE(s) returns TRUE, decoding fails.
   2.  Process s as follows:

   ss = s^2
   u1 = 1 - ss
   u2 = 1 + ss
   u2_sqr = u2^2

   v = -(D * u1^2) - u2_sqr

   (was_square, invsqrt) = SQRT_RATIO_M1(1, v * u2_sqr)

   den_x = invsqrt * u2
   den_y = invsqrt * den_x * v

   x = CT_ABS(2 * s * den_x)
   y = u1 * den_y
   t = x * y

   1.  If was_square is FALSE, IS_NEGATIVE(t) returns TRUE, or y = 0,
       decoding fails.  Otherwise, return the group element represented
       by the internal representation (x, y, 1, t) as the result of
       decoding.

```

### Code of Conduct

kesara:
Set `presentational_hints` option in WeasyPrint so that HTML
presentational hints are respected.

See https://doc.courtbouillon.org/weasyprint/stable/api_reference.html#supported-html-tags

Fixes #1054

### DIFF
--- a/tests/valid/docfile.html
+++ b/tests/valid/docfile.html
@@ -24,7 +24,7 @@
 <thead><tr>
 <td class="left"></td>
 <td class="center">Xml2rfc Vocabulary V3 Schema</td>
-<td class="right">October 2023</td>
+<td class="right">November 2023</td>
 </tr></thead>
 <tfoot><tr>
 <td class="left">xml2rfc(1)</td>
@@ -39,7 +39,7 @@
 <dd class="workgroup">xml2rfc(1)</dd>
 <dt class="label-published">Published:</dt>
 <dd class="published">
-<time datetime="2023-10-26" class="published">26 October 2023</time>
+<time datetime="2023-11-16" class="published">16 November 2023</time>
     </dd>
 <dt class="label-authors">Author:</dt>
 <dd class="authors">
--- a/tests/valid/draft-miek-test.html
+++ b/tests/valid/draft-miek-test.html
@@ -33,7 +33,7 @@
     intervaltree 3.1.0
     Jinja2 3.1.2
     lxml 4.9.3
-    platformdirs 3.11.0
+    platformdirs 3.10.0
     pycountry 22.3.5
     PyYAML 6.0.1
     requests 2.31.0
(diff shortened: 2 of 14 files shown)


rjsparks (approved):

kesara:
RFC 9xxx pdf outputs generated with this change: https://github.com/kesara/t4/tree/main/xml2rfc.pr1055.rfc9xxx

kesara:
Affected RFC8xxx PDFs: https://github.com/kesara/t4/tree/main/xml2rfc.pr1055.rfc8xxx.selected

PR #1092, opened 2024-01-31
fix: Silently pass when weasyprint is not installed

### LINKED ISSUE #1091: When weasyprint is not installed xml2rfc produces a log message
### Describe the issue

This issue is related to https://github.com/ietf-tools/xml2rfc/pull/1079

When `weasyprint` is not present, `xml2rfc` produces a log message:
On Python 3.11:
```
$ xml2rfc --text --html draft-foobar-00.xml
 Created file draft-foobar-00.txt
No package metadata was found for weasyprint
 Created file draft-foobar-00.html
```
or something like:
```
   xml2rfc draft-ietf-lsvr-l3dl.xml --html --text
     Created file draft-ietf-lsvr-l3dl.txt
    weasyprint
     Created file draft-ietf-lsvr-l3dl.html
```

### Code of Conduct

kesara:
Fixes #1091

### DIFF
--- a/xml2rfc/__init__.py
+++ b/xml2rfc/__init__.py
@@ -54,8 +54,7 @@ def get_versions():
             try:
                 dist = metadata.distribution(p)
                 versions.append((dist.metadata['name'], dist.metadata['version']))
-            except Exception as e:
-                print(e)
+            except:
                 pass
     except:
         pass


xml2rfc/__init__.py
```diff
@@ -54,8 +54,7 @@ def get_versions():
             try:
                 dist = metadata.distribution(p)
                 versions.append((dist.metadata['name'], dist.metadata['version']))
-            except Exception as e:
-                print(e)
+            except:
```
rjsparks, line 57: This will silently bury potentially more than "not installed". Is there a specific exception you could look for to silence the print rather than any exception?

kesara: This is used to get the versions of the dependencies. I think it's best to silently burry everything.

rjsparks (approved):

PR #1093, opened 2024-01-31
refactor: Use context managers when opening files

kesara:


jennifer-richards (approved):
lgtm

Off topic for the PR, but a couple of those might benefit from slight reorganization - instead of 
```python
  with open(...) as file:
    s = do_work()
    file.write(s)
```
you could refactor as
```python
  s = do_work()
  with open(...) as file:
    file.write(s)
```
and that will have the benefit that a failure in `do_work()` won't leave a file around. That'd make sense if such a failure is more likely than a problem opening the file.

Actually, there's also `pathlib.Path(...).write_text(...)` or `write_bytes(...)` that will do the latter without the `with` noise.

kesara:
> lgtm
> 
> Off topic for the PR, but a couple of those might benefit from slight reorganization - instead of
> 
> ```python
>   with open(...) as file:
>     s = do_work()
>     file.write(s)
> ```
> 
> you could refactor as
> 
> ```python
>   s = do_work()
>   with open(...) as file:
>     file.write(s)
> ```
> 
> and that will have the benefit that a failure in `do_work()` won't leave a file around. That'd make sense if such a failure is more likely than a problem opening the file.
> 

That's where context managers come in handy eh?
It will guarantee that the exit file will be closed regardless of failure happens during the execution of the `do_work()` or not.

> Actually, there's also `pathlib.Path(...).write_text(...)` or `write_bytes(...)` that will do the latter without the `with` noise.

jennifer-richards:
> It will guarantee that at the exit file will be closed regardless of failure happens during the execution of the `do_work()` or not.

Right, the file handle will be closed, but it'll leave behind an empty file. Whether you care about that is a separate question, I'm just pointing out that you could avoid creating it until you know that you have data to put in it.

kesara:
> > It will guarantee that at the exit file will be closed regardless of failure happens during the execution of the `do_work()` or not.
> 
> Right, the file handle will be closed, but it'll leave behind an empty file. Whether you care about that is a separate question, I'm just pointing out that you could avoid creating it until you know that you have data to put in it.

Good point. Yeah, this could be better.

PR #1104, opened 2024-02-15
fix(text): Don't break URLs in annotation

### LINKED ISSUE #1101: Extraneous space in eref within annotation in text output
### Describe the issue

In the txt output, an extraneous space has been added in eref within annotation of the [E4X] reference. This does not occur in the html or pdf outputs. 

   [E4X]      ISO, "Information technology - ECMAScript for XML (E4X)
              specification", Withdrawn, ISO/IEC 22537:2006, February
              2006, <https://www.iso.org/standard/41002.html>.  An
              equivalent specification, also withdrawn, can be obtained
              from <https://ecma-international.org/ publications-and-
              standards/standards/ecma-357>.

https://www.rfc-editor.org/v3test/rfc9535.txt
https://www.rfc-editor.org/v3test/rfc9535.html
https://www.rfc-editor.org/v3test/rfc9535.pdf
https://www.rfc-editor.org/v3test/rfc9535.xml

### Code of Conduct

kesara:
Fixes #1101

### DIFF
--- a/xml2rfc/writers/text.py
+++ b/xml2rfc/writers/text.py
@@ -37,6 +37,7 @@ from xml2rfc.utils import justify_inline, clean_text
 
 
 MAX_WIDTH = 72
+SPLITTER_WIDTH = 67
 
 IndexItem   = namedtuple('indexitem', ['item', 'subitem', 'anchor', 'page', ])
 Joiner      = namedtuple('joiner', ['join', 'indent', 'hang', 'overlap', 'do_outdent'])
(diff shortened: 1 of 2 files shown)


xml2rfc/writers/text.py
```diff
@@ -2746,7 +2749,8 @@ def render_reference(self, e, width, **kwargs):
         text += '.'
         for ctag in ('annotation', ):
             for c in e.iterdescendants(ctag):
-                text = self.tjoin(text, c, width, **kwargs)
+                # use MAX_WIDTH here since text gets formatted later
+                text = self.tjoin(text, c, MAX_WIDTH, keep_url=True, **kwargs)
```
rjsparks, line 2753: If we don't already have one, this deserves a test demonstrating what happens if we have annotation content that is larger than width but smaller than MAX_WIDTH going through this.

rjsparks (approved):

xml2rfc/writers/text.py
kesara: I've added unit tests with 881d8a17ddc226ca9f787eb152cd706b315d338a

rjsparks (approved):

PR #1106, opened 2024-02-19
fix: Allow non-ASCII values in all attributes

### LINKED ISSUE #1105: Error: Found non-ASCII content in <title> attribute value abbrev="...define short title... ≤ 40 chars"
### Describe the issue

This should not be an error.

### Code of Conduct

kesara:
PR #1017 allowed Unicord characters everywhere.
This fixes a bug that xml2rfc was guarding against non-ASCII characters in attribute values.
With `--warn-bare-unicode`, xml2rfc will warn if non-ASCII characters are present in attribute values.

Fixes #1105


cabo:
Thanks for the quick fix!

I'm not sure why you are taking out the latinscript tests for the ascii... attributes; I think these were useful.

cabo:
Looks good to me now.

The warning message uses "they" to refer to "non-ASCII content", maybe "it"?

rjsparks (approved):
Please include Carsten's grammar suggestion

PR #1107, opened 2024-02-21
ci: PyPI publish GHA should not run on push to main branch

kesara:


microamp (approved):
LGTM. I understand that you only want to publish the package via GHA manually.

:shipit:

PR #1134, opened 2024-07-01
feat: Remove dependency on six

### LINKED ISSUE #1130: Remove Python dependency six
### Description

`xml2rfc` doesn't support Python 2 but [six](https://github.com/benjaminp/six) has been used in several places.

### Code of Conduct

kesara:
* feat: Remove six. Fixes #1130 
* chore: Use context managers in `xml2rfc/parser.py`.

### DIFF
--- a/xml2rfc/parser.py
+++ b/xml2rfc/parser.py
@@ -12,7 +12,6 @@ import os
 import re
 import requests
 import shutil
-import six
 import time
 import xml2rfc.log
 import xml2rfc.utils
@@ -72,7 +71,7 @@ class CachingResolver(lxml.etree.Resolver):
 
         # Get directory of source
         if self.source:
-            if isinstance(self.source, six.string_types):
+            if isinstance(self.source, str):
                 self.source_dir = os.path.abspath(os.path.dirname(self.source))
             else:
                 self.source_dir = os.path.abspath(os.path.dirname(self.source.name))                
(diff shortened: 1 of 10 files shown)


xml2rfc/parser.py
```diff
@@ -543,47 +542,47 @@ def parse(self, remove_comments=True, remove_pis=False, quiet=False, strip_cdata
...
+            self.format_version = None
+            try:
+                for action, element in context:
+                    if element.tag == "rfc":
+                        self.rfc_number = element.attrib.get("number", None)
+                        self.format_version = element.attrib.get("version", None)
+                        break
+            except ValueError as e:
+                if e.message=="I/O operation on closed file":
+                    pass
```
microamp, line 585: Are we ignoring _all_ `ValueError`s here? Otherwise, 
```python
else:
    raise e
```
may be needed. Let me know if I'm reading it wrong.

microamp (approved):
Looks good to me. Just one minor comment.

xml2rfc/parser.py
kesara: Good question. It wasn't a change introduced by this PR.
But I don't think we should catch that error.
I'll remove the `try-except`.

kesara: Addressed with https://github.com/ietf-tools/xml2rfc/pull/1134/commits/b765594962d504d9d59181c4d69135e526959b50

microamp (approved):

PR #1137, opened 2024-07-09
fix(text): Render list items correctly

### LINKED ISSUE #1135: missing number in text output of <ol>
### Describe the issue

In Section 4.6 of RFC-to-be-9599, the `<ol>` tag contains two `<li>` tags, but the "2" for the second `<li>` tag does not appear in the [text output](https://www.rfc-editor.org/v3test/rfc9599.txt). 

Input:
input

Desired output:
desired output
(which is how it looks in the [HTML](https://www.rfc-editor.org/v3test/rfc9599.html) and [PDF](https://www.rfc-editor.org/v3test/rfc9599.pdf) outputs)

Actual output (not desired):
actual output (not desired)

Various workarounds have been tried by the author and the RPC without success.

### Code of Conduct

### LINKED ISSUE #1138: Missing style in <ul>
### Describe the issue

Similar to #1135, same bug is present in `ul`.
Example:
```
      <ul>
        <li>First Item</li>
        <li>
          <ul>
            <li>Second Item: Inner A</li>
            <li>Second Item: Inner B</li>
          </ul>
        </li>
        <li>Third Item</li>
      </ul>
```

Text output:
```
   *  First Item

      -  Second Item: Inner A

      -  Second Item: Inner B

   *  Third Item
```

### Code of Conduct

kesara:
Fixes #1135
Fixes #1138 

See changes made in https://github.com/ietf-tools/xml2rfc/commit/3712523cb325bb43b1b164e833dd436c1c62a295 for context.

This change can affect if `ol` is used inside a `li` as the first element and the `li` element is inside another `ol` element. 

I could not find any published RFCs with XML that match the above condition.

These are the RFCs that I found that use `ol` as the first element in `li`:
 * RFC8679
 * RFC8687

I didn't notice any changes in text output due to this change. But there are a few minor changes (with spacing) due to changes over the years.

### DIFF
--- a/tests/valid/elements.bom.text
+++ b/tests/valid/elements.bom.text
@@ -93,9 +93,9 @@ Table of Contents
        CMS References  . . . . . . . . . . . . . . . . . . . . . . .  25
        ESS References  . . . . . . . . . . . . . . . . . . . . . . .  26
        MIME-SPEC References  . . . . . . . . . . . . . . . . . . . .  26
-       Other References  . . . . . . . . . . . . . . . . . . . . . .  26
+       Other References  . . . . . . . . . . . . . . . . . . . . . .  27
    Appendix A.  Some Back Matter . . . . . . . . . . . . . . . . . .  27
-     A.1.  Contributors  . . . . . . . . . . . . . . . . . . . . . .  27
+     A.1.  Contributors  . . . . . . . . . . . . . . . . . . . . . .  28
      A.2.  Arbitrary^superscript in section _title_  . . . . . . . .  28
    Authors' Addresses  . . . . . . . . . . . . . . . . . . . . . . .  28
 
@@ -450,6 +450,7 @@ Author, et al.           Expires 13 January 2019                [Page 8]
 Internet-Draft       Xml2rfc Vocabulary V3 Elements            July 2018
 
 
+   4.
        4.1  Sublist with '%p' parent counter element in the list
             counter, item one
 
@@ -497,7 +498,6 @@ Internet-Draft       Xml2rfc Vocabulary V3 Elements            July 2018
 
       1.  Definition list, third bullet, fifth element: Ordered list,
           element li.  Text only.
-
 
 
 
@@ -1351,12 +1351,22 @@ Internet-Draft       Xml2rfc Vocabulary V3 Elements            July 2018
    *    Indent="5".  “The heaviest penalty for declining to rule is to
         be ruled by someone inferior to yourself.” Πολιτεία
 
+   *
                      Ich weiß nicht, was soll es bedeuten,
                      Daß ich so traurig bin;
                      Ein Mährchen aus alten Zeiten,
                      Das kommt mir nicht aus dem Sinn.
 
         Dreiunddreißig Gedichte
+
+   *    ul with no bullets
+
+           o  Must not have an empty line with bullet above this line.
+
+   *    ul with bullets
+
+        -
+           o  Must have an empty line with bullet above this line.
 
 References
 
@@ -1379,6 +1389,19 @@ Informative References
 
 CMS References
 
+
+
+
+
+
+
+
+
+Author, et al.           Expires 13 January 2019               [Page 25]
+
+Internet-Draft       Xml2rfc Vocabulary V3 Elements            July 2018
+
+
    [I-D.levkowetz-xml2rfc-v3-implementation-notes]
               Levkowetz, H., "Implementation notes for RFC7991,", Work
               in Progress, Internet-Draft, draft-levkowetz-xml2rfc-v3-
@@ -1394,13 +1417,6 @@ CMS References
    [RFC5652]  Housley, R., "Cryptographic Message Syntax (CMS)", STD 70,
               RFC 5652, DOI 10.17487/RFC5652, September 2009,
               <https://www.rfc-editor.org/info/rfc5652>.
-
-
-
-Author, et al.           Expires 13 January 2019               [Page 25]
-
-Internet-Draft       Xml2rfc Vocabulary V3 Elements            July 2018
-
 
 ESS References
 
@@ -1435,6 +1451,13 @@ MIME-SPEC References
               Examples", RFC 2049, DOI 10.17487/RFC2049, November 1996,
               <https://www.rfc-editor.org/info/rfc2049>.
 
+
+
+Author, et al.           Expires 13 January 2019               [Page 26]
+
+Internet-Draft       Xml2rfc Vocabulary V3 Elements            July 2018
+
+
    [RFC4289]  Freed, N. and J. Klensin, "Multipurpose Internet Mail
               Extensions (MIME) Part Four: Registration Procedures",
               BCP 13, RFC 4289, DOI 10.17487/RFC4289, December 2005,
@@ -1446,17 +1469,6 @@ MIME-SPEC References
               <https://www.rfc-editor.org/info/rfc6838>.
 
 Other References
-
-
-
-
-
-
-
-Author, et al.           Expires 13 January 2019               [Page 26]
-
-Internet-Draft       Xml2rfc Vocabulary V3 Elements            July 2018
-
 
    [AUTHORS]  Πυθαγόρας ὁ Σάμιος (Samos, P. O.), Doe, J., Bergström, A.,
               and J. Doe, "Lots of authors", RFC 7539,
@@ -1493,6 +1505,15 @@ Appendix A.  Some Back Matter
    // This is a long comment line which extends well beyond the
    // 72-character page width.
 
+
+
+
+
+Author, et al.           Expires 13 January 2019               [Page 27]
+
+Internet-Draft       Xml2rfc Vocabulary V3 Elements            July 2018
+
+
 A.1.  Contributors
 
    We'd like to thank these contributors:
@@ -1504,15 +1525,6 @@ A.1.  Contributors
    Israel
 
    Additional contact information:
-
-
-
-
-
-Author, et al.           Expires 13 January 2019               [Page 27]
-
-Internet-Draft       Xml2rfc Vocabulary V3 Elements            July 2018
-
 
       רוני אבן
       וואווי
@@ -1547,6 +1559,17 @@ Authors' Addresses
 
    Additional contact information:
 
+
+
+
+
+
+
+Author, et al.           Expires 13 January 2019               [Page 28]
+
+Internet-Draft       Xml2rfc Vocabulary V3 Elements            July 2018
+
+
       רוני אבן
       וואווי
       דוד המלך 14
@@ -1562,13 +1585,6 @@ Authors' Addresses
    Email: hanako.tanaka@nihon.example
 
    Additional contact information:
-
-
-
-Author, et al.           Expires 13 January 2019               [Page 28]
-
-Internet-Draft       Xml2rfc Vocabulary V3 Elements            July 2018
-
 
       田中花子 様
       日本
@@ -1600,22 +1616,6 @@ Internet-Draft       Xml2rfc Vocabulary V3 Elements            July 2018
    Sevedstorp 125
    Mariannelund
    Sweden
-
-
-
-
-
-
-
-
-
-
-
-
-
-
-
-
 
 
 
--- a/tests/valid/elements.text
+++ b/tests/valid/elements.text
@@ -377,6 +377,7 @@ Table of Contents
           Case Ambiguity [RFC8174], Appendix A.1 / Key Words Case
           Ambiguity [RFC8174] (Appendix A.1).
 
+   4.
        4.1  Sublist with '%p' parent counter element in the list
             counter, item one
 
@@ -1115,12 +1116,22 @@ Unnumbered section
    *    Indent="5".  “The heaviest penalty for declining to rule is to
         be ruled by someone inferior to yourself.” Πολιτεία
 
+   *
                      Ich weiß nicht, was soll es bedeuten,
                      Daß ich so traurig bin;
                      Ein Mährchen aus alten Zeiten,
                      Das kommt mir nicht aus dem Sinn.
 
         Dreiunddreißig Gedichte
+
+   *    ul with no bullets
+
+           o  Must not have an empty line with bullet above this line.
+
+   *    ul with bullets
+
+        -
+           o  Must have an empty line with bullet above this line.
 
 References
 
--- a/xml2rfc/writers/text.py
+++ b/xml2rfc/writers/text.py
@@ -2152,12 +2152,21 @@ class TextWriter(BaseV3Writer):
         tt, __ = self.text_or_block_renderer(e, width, **kwargs)
         if isinstance(tt, list):
             lines = stripl(tt)
+            children = e.getchildren()
             if (
-                lines
-                and lines[0].elem.tag not in ["artwork", "figure", "sourcecode"]
-                and not (p.tag == "ul" and lines[0].elem.tag == "li")
+                children
+                and children[0].tag in ["artwork", "figure", "sourcecode", "ol", "ul"]
+                and not (p.tag == "ul" and p.get("empty", "false").lower() == "true")
             ):
-                lines[0].text = text + lines[0].text.lstrip(stripspace)
+                 lines.insert(0, Line(text.rstrip(stripspace), e))
+            else:
+                if lines and lines[0].elem.tag not in [
+                    "artwork",
+                    "figure",
+                    "sourcecode",
+                    "li",
+                ]:
+                    lines[0].text = text + lines[0].text.lstrip(stripspace)
         else:
             text += tt.lstrip(stripspace)
             lines = mklines(text, e)
--- a/xml2rfc/writers/text.py
+++ b/xml2rfc/writers/text.py
@@ -2152,7 +2152,11 @@ class TextWriter(BaseV3Writer):
         tt, __ = self.text_or_block_renderer(e, width, **kwargs)
         if isinstance(tt, list):
             lines = stripl(tt)
-            if lines and lines[0].elem.tag not in ['artwork', 'figure', 'sourcecode', 'li', ]:
+            if (
+                lines
+                and lines[0].elem.tag not in ["artwork", "figure", "sourcecode"]
+                and not (p.tag == "ul" and lines[0].elem.tag == "li")
+            ):
                 lines[0].text = text + lines[0].text.lstrip(stripspace)
         else:
             text += tt.lstrip(stripspace)
--- a/tests/input/elements.xml
+++ b/tests/input/elements.xml
@@ -1362,6 +1362,26 @@ for opt, value in opts:
                </artwork>
                <t><xref target="GEDICHTE" format="title"/></t>
             </li>
+            <li>
+              <t>ul with no bullets</t>
+              <ul empty="true">
+                <li>
+                  <ul>
+                    <li>Must not have an empty line with bullet above this line.</li>
+                  </ul>
+                </li>
+              </ul>
+            </li>
+            <li>
+              <t>ul with bullets</t>
+              <ul>
+                <li>
+                  <ul>
+                    <li>Must have an empty line with bullet above this line.</li>
+                  </ul>
+                </li>
+              </ul>
+            </li>
          </ul>
       </section>
    </middle>
--- a/tests/valid/elements.pages.text
+++ b/tests/valid/elements.pages.text
@@ -93,9 +93,9 @@ Table of Contents
        CMS References  . . . . . . . . . . . . . . . . . . . . . . .  25
        ESS References  . . . . . . . . . . . . . . . . . . . . . . .  26
        MIME-SPEC References  . . . . . . . . . . . . . . . . . . . .  26
-       Other References  . . . . . . . . . . . . . . . . . . . . . .  26
+       Other References  . . . . . . . . . . . . . . . . . . . . . .  27
    Appendix A.  Some Back Matter . . . . . . . . . . . . . . . . . .  27
-     A.1.  Contributors  . . . . . . . . . . . . . . . . . . . . . .  27
+     A.1.  Contributors  . . . . . . . . . . . . . . . . . . . . . .  28
      A.2.  Arbitrary^superscript in section _title_  . . . . . . . .  28
    Authors' Addresses  . . . . . . . . . . . . . . . . . . . . . . .  28
 
@@ -450,6 +450,7 @@ Author, et al.          Expires January 13, 2019                [Page 8]
 Internet-Draft       Xml2rfc Vocabulary V3 Elements            July 2018
 
 
+   4.
        4.1  Sublist with '%p' parent counter element in the list
             counter, item one
 
@@ -497,7 +498,6 @@ Internet-Draft       Xml2rfc Vocabulary V3 Elements            July 2018
 
       1.  Definition list, third bullet, fifth element: Ordered list,
           element li.  Text only.
-
 
 
 
@@ -1351,12 +1351,22 @@ Internet-Draft       Xml2rfc Vocabulary V3 Elements            July 2018
    *    Indent="5".  “The heaviest penalty for declining to rule is to
         be ruled by someone inferior to yourself.” Πολιτεία
 
+   *
                      Ich weiß nicht, was soll es bedeuten,
                      Daß ich so traurig bin;
                      Ein Mährchen aus alten Zeiten,
                      Das kommt mir nicht aus dem Sinn.
 
         Dreiunddreißig Gedichte
+
+   *    ul with no bullets
+
+           o  Must not have an empty line with bullet above this line.
+
+   *    ul with bullets
+
+        -
+           o  Must have an empty line with bullet above this line.
 
 References
 
@@ -1379,6 +1389,19 @@ Informative References
 
 CMS References
 
+
+
+
+
+
+
+
+
+Author, et al.          Expires January 13, 2019               [Page 25]
+
+Internet-Draft       Xml2rfc Vocabulary V3 Elements            July 2018
+
+
    [I-D.levkowetz-xml2rfc-v3-implementation-notes]
               Levkowetz, H., "Implementation notes for RFC7991,", Work
               in Progress, Internet-Draft, draft-levkowetz-xml2rfc-v3-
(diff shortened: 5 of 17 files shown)


tests/valid/elements.bom.text
```diff
@@ -450,7 +450,7 @@ Author, et al.           Expires 13 January 2019                [Page 8]
 Internet-Draft       Xml2rfc Vocabulary V3 Elements            July 2018
 
 
-       4.1  Sublist with '%p' parent counter element in the list
+   4.  4.1  Sublist with '%p' parent counter element in the list
```
rjsparks, line 453: This is unexpected I think.

xml2rfc/writers/text.py
```diff
@@ -2150,7 +2150,9 @@ def render_li(self, e, width, **kwargs):
         tt, __ = self.text_or_block_renderer(e, width, **kwargs)
         if isinstance(tt, list):
             lines = stripl(tt)
-            if lines and lines[0].elem.tag not in ['artwork', 'figure', 'sourcecode', 'li', ]:
+            if (lines and
```
rjsparks, line 2153: is this what black would produce?

kesara: Updated with https://github.com/ietf-tools/xml2rfc/pull/1137/commits/fdb6fb8ad2024b3483f7480a07f8370e673c27e2

tests/valid/elements.bom.text
kesara: This is the expected outcome of this change.
But yes, this would be an unexpected for a draft output's original intention of the author because it relies on the `xml2rfc` *bug*.
As mentioned in the description, I didn't find any published RFCs (in XML era) affected by this change.
I haven't done such a search on I-Ds but the chances of this being exploited should be high.

If we don't want to fix this, there are workarounds for the RFC-to-be [^1][^2].

[^1]: https://github.com/ietf-tools/xml2rfc/issues/1135#issuecomment-2216307530
[^2]: https://github.com/ietf-tools/xml2rfc/issues/1135#issuecomment-2217025945

rjsparks:
I think it's worth trying to change things such that generatiing
```
   4.  4.1  Sublist with '%p' parent counter element in the list
            counter, item one

       4.2  Item two.

       4.3  Item three.
```

we generate
```
   4. 

       4.1  Sublist with '%p' parent counter element in the list
            counter, item one

       4.2  Item two.

       4.3  Item three.
```

Unless that breaks the intended fix for the rfc-to-be.

cabo:
What is better about that superfluous empty line?

(You can probably have one if you want with `<br>`)

rjsparks:
Look at it in context with the rest of the nested `<ol>` above it in the test document:  https://github.com/ietf-tools/xml2rfc/blob/fdb6fb8ad2024b3483f7480a07f8370e673c27e2/tests/valid/elements.bom.text#L366-L458 and consider what would happen if there were any text between the <ol> and <li> at https://github.com/ietf-tools/xml2rfc/blob/fdb6fb8ad2024b3483f7480a07f8370e673c27e2/tests/input/elements.xml#L472-L473

kesara:
I have converted PR to a draft.
The changes from https://github.com/ietf-tools/xml2rfc/pull/1137/commits/e889f930b982c9bc9045f8a6e653f25a6464ec58 add a new line when with the **list style** to both `ol` and `ul`.
This was skipped previously.
Specifically, this adds **list style** + new line when `li` contains an `artwork`, `figure`, or `sourcecode` as the first child.
See how this changed test outputs in https://github.com/ietf-tools/xml2rfc/pull/1137/commits/ca8286a4126bc28bba095eb7e708c5a7424aadef
Note that the test output has date changes and multiple content changes propagated because of the addition of new lines.

tests/valid/elements.text
```diff
@@ -1092,6 +1093,7 @@ Unnumbered section
    *    Indent="5".  “The heaviest penalty for declining to rule is to
         be ruled by someone inferior to yourself.” Πολιτεία
 
+   *
```
rjsparks, line 1096: And now this is perhaps unfortunate. What we've uncovered is, I think, a decision to special case the rendering of nested lists without leading text that doesn't generalize to all possible uses of such things, and we're just toggling cross the edges of that earlier decision.

Has the previous side-effect of no leading marker been used in _this_ case previously? Is it ok to make this change (since getting the behavior of that side-effect is possible in better defined ways, I _think)?

@cabo - what do you think the right thing to do with this is?

kesara: Here's a list of RFCs that has `ul` tag inside a `li` as first element:
```
rfc8668.xml
rfc8742.xml
rfc8811.xml
rfc8793.xml
rfc8927.xml
rfc8677.xml
rfc8667.xml
rfc8698.xml
rfc9195.xml
```
All these use `empty="true"` in parent `ul`, so no new `*` gets added but there's a bug in the suggested fix because it adds a new line. I'll fix that.

cabo: > @cabo - what do you think the right thing to do with this is?

I don't think it is ever correct to add a new line that the author didn't want.
If it is needed, the author can put it in, but it never should be forced on them.

kesara: I have pushed a fix to handle `empty="true"`.
See https://github.com/ietf-tools/xml2rfc/pull/1137/commits/11ef6ac0d304a1ee204ae3325e3d3ca700cfc546 and https://github.com/ietf-tools/xml2rfc/pull/1137/commits/fe595b1376d9359f27a3879b633d272d4d016f45

kesara: @rjsparks & @cabo For context HTML output doesn't add a new line.
Screenshot 2024-07-12 at 19 49 28

rjsparks: And HTML has the marker, so the text should to - that convinces me that (except for the newline that Carsten doesn't want) the current proposed change is right. 

So people looking at this later can see it in context, could you also paste the section of html analogous to  the `4.  4.1` change in the text in this thread?

kesara: @rjsparks `ol` has a new line in HTML:
Screenshot 2024-07-13 at 14 23 39

kesara: @rjsparks The above screenshot is from Firefox and that might not be showing the correct rending. See #1158

kesara:
@rjsparks
Here's list of affected[^1] published RFC diffs:
* https://author-tools.ietf.org/diff?url_1=https://github.com/kesara/t4/raw/refs/heads/main/xml2rfc-1137/rfc8742.txt
* https://author-tools.ietf.org/diff?url_1=https://github.com/kesara/t4/raw/refs/heads/main/xml2rfc-1137/rfc8811.txt
* https://author-tools.ietf.org/diff?url_1=https://github.com/kesara/t4/raw/refs/heads/main/xml2rfc-1137/rfc8793.txt
* https://author-tools.ietf.org/diff?url_1=https://github.com/kesara/t4/raw/refs/heads/main/xml2rfc-1137/rfc8668.txt
* https://author-tools.ietf.org/diff?url_1=https://github.com/kesara/t4/raw/refs/heads/main/xml2rfc-1137/rfc8687.txt
* https://author-tools.ietf.org/diff?url_1=https://github.com/kesara/t4/raw/refs/heads/main/xml2rfc-1137/rfc9599.txt
* https://author-tools.ietf.org/diff?url_1=https://github.com/kesara/t4/raw/refs/heads/main/xml2rfc-1137/rfc8679.txt
* https://author-tools.ietf.org/diff?url_1=https://github.com/kesara/t4/raw/refs/heads/main/xml2rfc-1137/rfc8927.txt
* https://author-tools.ietf.org/diff?url_1=https://github.com/kesara/t4/raw/refs/heads/main/xml2rfc-1137/rfc8677.txt
* https://author-tools.ietf.org/diff?url_1=https://github.com/kesara/t4/raw/refs/heads/main/xml2rfc-1137/rfc8667.txt
* https://author-tools.ietf.org/diff?url_1=https://github.com/kesara/t4/raw/refs/heads/main/xml2rfc-1137/rfc8698.txt
* https://author-tools.ietf.org/diff?url_1=https://github.com/kesara/t4/raw/refs/heads/main/xml2rfc-1137/rfc9195.txt
* https://author-tools.ietf.org/diff?url_1=https://github.com/kesara/t4/raw/refs/heads/main/xml2rfc-1137/rfc9687.txt

[^1]: https://t4.fq.nz/xml2rfc-1137/output.txt

rjsparks:
The above list is a bit confusing? I only see impact from this PRs change on rfc9599.

kesara:
> The above list is a bit confusing? I only see impact from this PRs change on rfc9599.

In most case there's `empty="true"` attribute set on the element or the parent level, hence no changes.

kesara:
List of *active* I-Ds that match the crieria[^1]:

- https://author-tools.ietf.org/diff?url_1=https://www.ietf.org/archive/id/draft-bernardos-cats-anchoring-service-mobility-01.txt&url_2=https://github.com/kesara/t4/raw/main/xml2rfc-1137-ids/draft-bernardos-cats-anchoring-service-mobility-01.txt
- https://author-tools.ietf.org/diff?url_1=https://www.ietf.org/archive/id/draft-bernardos-cats-ip-address-anchoring-02.txt&url_2=https://github.com/kesara/t4/raw/main/xml2rfc-1137-ids/draft-bernardos-cats-ip-address-anchoring-02.txt
- https://author-tools.ietf.org/diff?url_1=https://www.ietf.org/archive/id/draft-davis-nmop-some-refinements-to-rfc8345-00.txt&url_2=https://github.com/kesara/t4/raw/main/xml2rfc-1137-ids/draft-davis-nmop-some-refinements-to-rfc8345-00.txt
- https://author-tools.ietf.org/diff?url_1=https://www.ietf.org/archive/id/draft-dutt-nvo3-rfc7348bis-00.txt&url_2=https://github.com/kesara/t4/raw/main/xml2rfc-1137-ids/draft-dutt-nvo3-rfc7348bis-00.txt
- https://author-tools.ietf.org/diff?url_1=https://www.ietf.org/archive/id/draft-haynes-nfsv4-flexfiles-v2-00.txt&url_2=https://github.com/kesara/t4/raw/main/xml2rfc-1137-ids/draft-haynes-nfsv4-flexfiles-v2-00.txt
- https://author-tools.ietf.org/diff?url_1=https://www.ietf.org/archive/id/draft-ietf-bess-evpn-fast-df-recovery-12.txt&url_2=https://github.com/kesara/t4/raw/main/xml2rfc-1137-ids/draft-ietf-bess-evpn-fast-df-recovery-12.txt
- https://author-tools.ietf.org/diff?url_1=https://www.ietf.org/archive/id/draft-ietf-bess-rfc7432bis-10.txt&url_2=https://github.com/kesara/t4/raw/main/xml2rfc-1137-ids/draft-ietf-bess-rfc7432bis-10.txt
- https://author-tools.ietf.org/diff?url_1=https://www.ietf.org/archive/id/draft-ietf-dnssd-srp-25.txt&url_2=https://github.com/kesara/t4/raw/main/xml2rfc-1137-ids/draft-ietf-dnssd-srp-25.txt
- https://author-tools.ietf.org/diff?url_1=https://www.ietf.org/archive/id/draft-ietf-drip-dki-03.txt&url_2=https://github.com/kesara/t4/raw/main/xml2rfc-1137-ids/draft-ietf-drip-dki-03.txt
- https://author-tools.ietf.org/diff?url_1=https://www.ietf.org/archive/id/draft-ietf-emu-eap-fido-00.txt&url_2=https://github.com/kesara/t4/raw/main/xml2rfc-1137-ids/draft-ietf-emu-eap-fido-00.txt
- https://author-tools.ietf.org/diff?url_1=https://www.ietf.org/archive/id/draft-ietf-idr-bgp-generic-metric-00.txt&url_2=https://github.com/kesara/t4/raw/main/xml2rfc-1137-ids/draft-ietf-idr-bgp-generic-metric-00.txt
- https://author-tools.ietf.org/diff?url_1=https://www.ietf.org/archive/id/draft-ietf-lamps-rfc5272bis-01.txt&url_2=https://github.com/kesara/t4/raw/main/xml2rfc-1137-ids/draft-ietf-lamps-rfc5272bis-01.txt
- https://author-tools.ietf.org/diff?url_1=https://www.ietf.org/archive/id/draft-ietf-pce-flexible-grid-11.txt&url_2=https://github.com/kesara/t4/raw/main/xml2rfc-1137-ids/draft-ietf-pce-flexible-grid-11.txt
- https://author-tools.ietf.org/diff?url_1=https://www.ietf.org/archive/id/draft-ietf-radext-deprecating-radius-05.txt&url_2=https://github.com/kesara/t4/raw/main/xml2rfc-1137-ids/draft-ietf-radext-deprecating-radius-05.txt
- https://author-tools.ietf.org/diff?url_1=https://www.ietf.org/archive/id/draft-ietf-radext-tls-psk-11.txt&url_2=https://github.com/kesara/t4/raw/main/xml2rfc-1137-ids/draft-ietf-radext-tls-psk-11.txt
- https://author-tools.ietf.org/diff?url_1=https://www.ietf.org/archive/id/draft-ietf-rats-corim-06.txt&url_2=https://github.com/kesara/t4/raw/main/xml2rfc-1137-ids/draft-ietf-rats-corim-06.txt
- https://author-tools.ietf.org/diff?url_1=https://www.ietf.org/archive/id/draft-ietf-rift-rift-24.txt&url_2=https://github.com/kesara/t4/raw/main/xml2rfc-1137-ids/draft-ietf-rift-rift-24.txt
- https://author-tools.ietf.org/diff?url_1=https://www.ietf.org/archive/id/draft-ietf-sidrops-aspa-slurm-02.txt&url_2=https://github.com/kesara/t4/raw/main/xml2rfc-1137-ids/draft-ietf-sidrops-aspa-slurm-02.txt
- https://author-tools.ietf.org/diff?url_1=https://www.ietf.org/archive/id/draft-ietf-snac-simple-06.txt&url_2=https://github.com/kesara/t4/raw/main/xml2rfc-1137-ids/draft-ietf-snac-simple-06.txt
- https://author-tools.ietf.org/diff?url_1=https://www.ietf.org/archive/id/draft-ietf-teas-nrp-scalability-06.txt&url_2=https://github.com/kesara/t4/raw/main/xml2rfc-1137-ids/draft-ietf-teas-nrp-scalability-06.txt
- https://author-tools.ietf.org/diff?url_1=https://www.ietf.org/archive/id/draft-ietf-teas-yang-te-37.txt&url_2=https://github.com/kesara/t4/raw/main/xml2rfc-1137-ids/draft-ietf-teas-yang-te-37.txt
- https://author-tools.ietf.org/diff?url_1=https://www.ietf.org/archive/id/draft-ietf-tls-wkech-06.txt&url_2=https://github.com/kesara/t4/raw/main/xml2rfc-1137-ids/draft-ietf-tls-wkech-06.txt
- https://author-tools.ietf.org/diff?url_1=https://www.ietf.org/archive/id/draft-lim-apv-03.txt&url_2=https://github.com/kesara/t4/raw/main/xml2rfc-1137-ids/draft-lim-apv-03.txt
- https://author-tools.ietf.org/diff?url_1=https://www.ietf.org/archive/id/draft-lim-rtp-apv-00.txt&url_2=https://github.com/kesara/t4/raw/main/xml2rfc-1137-ids/draft-lim-rtp-apv-00.txt
- https://author-tools.ietf.org/diff?url_1=https://www.ietf.org/archive/id/draft-sajassi-bess-rfc8317bis-03.txt&url_2=https://github.com/kesara/t4/raw/main/xml2rfc-1137-ids/draft-sajassi-bess-rfc8317bis-03.txt
- https://author-tools.ietf.org/diff?url_1=https://www.ietf.org/archive/id/draft-schanzen-r5n-06.txt&url_2=https://github.com/kesara/t4/raw/main/xml2rfc-1137-ids/draft-schanzen-r5n-06.txt
- https://author-tools.ietf.org/diff?url_1=https://www.ietf.org/archive/id/draft-schmutzer-bess-bitstream-vpws-signalling-02.txt&url_2=https://github.com/kesara/t4/raw/main/xml2rfc-1137-ids/draft-schmutzer-bess-bitstream-vpws-signalling-02.txt
- https://author-tools.ietf.org/diff?url_1=https://www.ietf.org/archive/id/draft-schmutzer-pals-ple-signaling-02.txt&url_2=https://github.com/kesara/t4/raw/main/xml2rfc-1137-ids/draft-schmutzer-pals-ple-signaling-02.txt
- https://author-tools.ietf.org/diff?url_1=https://www.ietf.org/archive/id/draft-spaghetti-sidrops-aspa-slurm-03.txt&url_2=https://github.com/kesara/t4/raw/main/xml2rfc-1137-ids/draft-spaghetti-sidrops-aspa-slurm-03.txt
- https://author-tools.ietf.org/diff?url_1=https://www.ietf.org/archive/id/draft-tomas-openroaming-04.txt&url_2=https://github.com/kesara/t4/raw/main/xml2rfc-1137-ids/draft-tomas-openroaming-04.txt
- https://author-tools.ietf.org/diff?url_1=https://www.ietf.org/archive/id/draft-yan-dmm-man-14.txt&url_2=https://github.com/kesara/t4/raw/main/xml2rfc-1137-ids/draft-yan-dmm-man-14.txt

[^1]: https://t4.fq.nz/xml2rfc-1137-ids/i-d-output.txt

kesara:
@rjsparks, On that list I only found following I-D to have a output change because of the proposed fix: https://author-tools.ietf.org/diff?url_1=https://www.ietf.org/archive/id/draft-haynes-nfsv4-flexfiles-v2-00.txt&url_2=https://github.com/kesara/t4/raw/main/xml2rfc-1137-ids/draft-haynes-nfsv4-flexfiles-v2-00.txt

rjsparks:
Is it clear what the authors of that draft could do to get the output they originally wanted given this change?
(The source for that draft appears to be at https://github.com/loghyr/flexfilesv2)

kesara:
> Is it clear what the authors of that draft could do to get the output they originally wanted given this change? (The source for that draft appears to be at https://github.com/loghyr/flexfilesv2)

@rjsparks The original document with the issue RFC9599 has already published with the suggested workaround using `<t>`. Note this is RFC9599's text output will change slightly with this PR[^1].

This is the changes for latest draft from https://github.com/loghyr/flexfilesv2:
Screenshot 2025-01-10 at 18 38 31


[^1]: https://author-tools.ietf.org/diff?url_1=https://github.com/kesara/t4/raw/refs/heads/main/xml2rfc-1137/rfc9599.txt

rjsparks:
so this would leave flexfilesv2 unchanged with this PR applied then:
```
        <li>
          <t>Delegations are assigned by the metadata server that initiates
          recalls when conflicting OPENs are processed.  Because I/O
          operations are allowed to present delegation stateids, the
          metadata server requires the ability:</t>
          <ol>
            <li>
              to make the storage device aware of the association between
              the metadata-server-chosen stateid and the filehandle and
              delegation type it represents
            </li>

            <li>
              to break such an association.
            </li>
          </ol>
        </li>
```

kesara:
@rjsparks yes, that will have unchanged output.

ajeanmahoney (commented):
See my comments on the PR:
https://github.com/ietf-tools/xml2rfc/pull/1137#issuecomment-2588377565

cabo:
> There were several diffs listed above that show changes to the list structure that I think the authors would not want:

I agree.
The output with these changes no longer reflects the list structure; a new item is created in the outer list just because one list item in there has an inner list.

kesara:
> RFC 9599 kicked off this issue. This proposed update causes the text output not to match the HTML and PDF outputs as well as the original workaround did: https://author-tools.ietf.org/diff?url_1=https://github.com/kesara/t4/raw/refs/heads/main/xml2rfc-1137/rfc9599.txt
> 
> There were several diffs listed above that show changes to the list structure that I think the authors would not want:
> 
>     * https://author-tools.ietf.org/diff?url_1=https://www.ietf.org/archive/id/draft-haynes-nfsv4-flexfiles-v2-00.txt&url_2=https://github.com/kesara/t4/raw/main/xml2rfc-1137-ids/draft-haynes-nfsv4-flexfiles-v2-00.txt
> 
>     * https://author-tools.ietf.org/diff?url_1=https://www.ietf.org/archive/id/draft-ietf-dnssd-srp-25.txt&url_2=https://github.com/kesara/t4/raw/main/xml2rfc-1137-ids/draft-ietf-dnssd-srp-25.txt
> 
>     * https://author-tools.ietf.org/diff?url_1=https://www.ietf.org/archive/id/draft-ietf-rift-rift-24.txt&url_2=https://github.com/kesara/t4/raw/main/xml2rfc-1137-ids/draft-ietf-rift-rift-24.txt
> 
>     * https://author-tools.ietf.org/diff?url_1=https://www.ietf.org/archive/id/draft-ietf-snac-simple-06.txt&url_2=https://github.com/kesara/t4/raw/main/xml2rfc-1137-ids/draft-ietf-snac-simple-06.txt
> 
>     * https://author-tools.ietf.org/diff?url_1=https://www.ietf.org/archive/id/draft-schanzen-r5n-06.txt&url_2=https://github.com/kesara/t4/raw/main/xml2rfc-1137-ids/draft-schanzen-r5n-06.txt
> 
> 
> IIUC, lists with empty="true" don't have changes in the list structure. However, the RPC evaluates the use of empty="true", and we may want to apply a more appropriate tag (see https://www.rfc-editor.org/pubprocess/how-we-update/). Would need to keep empty="true" to avoid some of these unwanted list changes?

@ajeanmahoney `emptry="true"` will only work for `ul` and what it does is not setting the `li` bullets empty. I don't see how that can avoid any changes from this PR because it's set a parent level.

kesara:
@rjsparks, @ajeanmahoney, @cabo, I think we have 3 alternatives here:

1. Keep the empty line. (Current PR does this)
2. Don't keep an empty line but add the missing list style (bullet/number/char).
3. Don't do anything and leave the bug as it-is.

BTW `xml2rfc` will probably have to add an empty line with the list style, when `li` contains an `artwork`, `figure`, `sourcecode` etc. as the first child to avoid confusion.

cabo:
> @rjsparks, @ajeanmahoney, @cabo, I think we have 3 alternatives here:
> 
> 	• Keep the empty line. (Current PR does this)

I’m not sure I know which line is empty, but maybe you are talking about the line with just a bullet that is being added?
The bullet wasn’t in the source, so it shouldn’t be in the output.

> 	• Don't keep an empty line but add the missing list style (bullet/number/char).

The bullet wasn’t in the source, so it shouldn’t be in the output.
There should be exactly as many bullets as there are <li elements.

> 	• Don't do anything and leave the bug as it-is.

(I need to look up what the bug was…)

> BTW xml2rfc will probably have to add an empty line with the list style, when li contains an artwork, figure, sourcecode etc. as the first child to avoid confusion.

I’m not sure what the problem is

1. int main() {}
2. public static void main(String args[])

Sourcecode and artwork can be in exactly the same place where normal text can be, no newline needed.

Grüße, Carsten

rjsparks:
@cabo: from the beginning of the PR:
Fixes https://github.com/ietf-tools/xml2rfc/issues/1135
Fixes https://github.com/ietf-tools/xml2rfc/issues/1138

Note in the original report that the txt output is currently fundamentally different from the html/pdf output. If there's something to address here, it's _that_.

The original issue can be addressed the same way I suggested the flexfilesv2 doc could be repaired above, and maybe the outcome of this whole thread is "If it hurts, don't do it" with advice to put the lead in a `<t>` (or an empty `<t>` if a nested list is really a new item in the outer list. But that would leave the current issue where the txt output has a significant variance from the html unaddressed.

alicerusso:
@rjsparks wrote:
> maybe the outcome of this whole thread is "If it hurts, don't do it" with advice to put the lead in a <t> (or an empty <t> if a nested list is really a new item in the outer list.

Agree.

> But that would leave the current issue where the txt output has a significant variance from the html unaddressed.

I wonder about giving a warning so that the user is aware that the input yields this variance in the output files.

cabo:
The examples in the referenced issues all show entirely reasonable text structures that for some reason aren't show correctly in the plaintext form.
Nothing hurts except for the bug.
Please fix the bug.

cabo:
> This is the changes for latest draft from https://github.com/loghyr/flexfilesv2:

Well, that example is bad because the text structure doesn't reflect the text content.
Indeed, the author here should remove the superfluous <li>.
The renderer cannot get this right, GIGO.

ajeanmahoney:
> @rjsparks wrote:
>
>> maybe the outcome of this whole thread is "If it hurts, don't do it" with advice to put the lead in a (or an empty if a nested list is really a new item in the outer list.
>
>@alicerusso Agree.  

Also agree. I would have suggested a lead-in sentence to the authors as a workaround, but the workaround suggested by @kesara in #1135 also works. This is an obscure corner case for nested lists that isn't covered by the Chicago Manual of Style, which says to use your word processor's outline list functionality to format lists correctly. FWIW I tried to get MS Word and Google Docs to recreate the following text from RFC 9599 and couldn't:
```
   1.  approximate preservation of the presence (and therefore timing)
       of congestion marks on the L2 frames used to construct an IP
       packet;

   2.  a.  at high frequency of congestion marking, approximate
           preservation of the proportion of congestion marks arriving
           and departing;

       b.  at low frequency of congestion marking, approximate
           preservation of the timing of congestion marks arriving and
           departing.
``` 
I'm good with the workaround here.

cabo:
> 
>    1.  approximate preservation of the presence (and therefore timing)
>        of congestion marks on the L2 frames used to construct an IP
>        packet;
> 
>    2.  a.  at high frequency of congestion marking, approximate
>            preservation of the proportion of congestion marks arriving
>            and departing;
> 
>        b.  at low frequency of congestion marking, approximate
>            preservation of the timing of congestion marks arriving and
>            departing.

First of all, people should be able to create exactly this nested list if it is their intention.

This is an <ol with two <li, where the second <li contains an <ol with two <li.

I have a hard time generating respect for a text formatting system that can’t do nested lists this trivial way.

Note that

> 
>    1.  approximate preservation of the presence (and therefore timing)
>        of congestion marks on the L2 frames used to construct an IP
>        packet;
> 
>        a.  at high frequency of congestion marking, approximate
>            preservation of the proportion of congestion marks arriving
>            and departing;
> 
>        b.  at low frequency of congestion marking, approximate
>            preservation of the timing of congestion marks arriving and
>            departing.

…also is a nested list that authors might want to create.

This is an <ol with one <li, which has a paragraph and an <ol with two <li as content.

Grüße, Carsten

ajeanmahoney:
@cabo  Authors can create such a list with the formatting described here:
https://github.com/ietf-tools/xml2rfc/issues/1135#issuecomment-2216307530

I've called it a workaround, but it's not really. Using `<t/>` within `<li/>` isn't a hack.

cabo:
> Using `<t/>` within `<li/>` isn't a hack.

Well, allowing both mixed content (span-level) and  t (block-level) sure is, avoiding the h word, interesting.
Giving them different semantics makes my mind boggle.

One problem here is that in markdown there is no difference, which is pretty much the reason why kramdown-rfc is still generating v2 for these because the v2v3 converter seems to do something that generally works.
(But not always.)

One other problem is that the content models of <t> and <li> are different, which creates serious headaches.

rjsparks:
@cabo 's last comment is about the TBD RSWG(?) work on normalizing the content model for the whole grammar.

I think we close this particular issue as WONTFIX (specifically, we don't merge this proposed code change).

kesara:
Closing this as a WONTFIX.

PR #1140, opened 2024-07-10
fix(text): Update non-ASCII art text

### LINKED ISSUE #617: use rfc filename in text output when only svg is provided
`type_defect`   |    by arusso@amsl.com

___

In the text output of an RFC, when there is artwork type="svg" provided, and no corresponding artwork type="ascii-art" (i.e., artset is not used), the rfc@docName appears in the message in the text output. Seems it should be the RFC number.

Current output:
```
      (Artwork only available as svg: No external link available, see
      draft-carpenter-eligibility-expand-10.html for artwork.)
```

Desired output (perhaps "No external link available" is not necessary, and "for artwork" is redundant):
```
      (Artwork only available as SVG: see rfc8989.html)
```

Or provide a URL for ease of use:
```
      (Artwork only available as SVG: see 
      https://www.rfc-editor.org/rfc/rfc8989.html)
```

Files (see Figure 1):
https://www.rfc-editor.org/v3test/test8989.txt
https://www.rfc-editor.org/v3test/test8989.xml
https://www.rfc-editor.org/v3test/test8989.html

[Note: The test file is a modification of RFC 8989; artwork was removed. RFC 8989 does contain the "ascii-art" equivalent and uses <artset> accordingly.]

___
_Issue migrated from trac:617 at 2022-02-08 07:14:03 +0000_

kesara:
Fixes #617

This change introduces two new command line options for text:
```
       --rfc-html-archive-url
          Default value: https://www.rfc-editor.org/rfc/
          URL for HTML file archive of RFCs.

       --id-html-archive-url
          Default value: https://www.ietf.org/archive/id/
          URL for HTML file archive Internet-Drafts.
```

##### Update text outputs:
For I-Ds:
```
(Artwork only available as svg: see https://www.ietf.org/archive/id/<docname>.html)
```

For RFCs:
```
(Artwork only available as svg: see https://www.rfc-editor.org/rfc/rfc<rfc number>.html)
```


cabo:
Is it possible to link to the actual figure?

kesara:
@cabo `figure` for `artwork` is optional. HTML processor adds anchors when there is no `figure` for artwork. HTML processing and text processing are two separate processors.
It is hard to guess the HTML anchors by text processor accurately.

rjsparks (approved):

PR #1186, opened 2024-11-20
chore: Cleanup MANIFEST.in

kesara:
Remove obsolete files and tests files from the MANIFEST.in

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,12 +1,6 @@
 include LICENSE
 include CHANGELOG.md
 include README.md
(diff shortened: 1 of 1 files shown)


rjsparks:
There's a package philosophy change being made here from "a pip install gives you a dev environment" to "if you want a dev environment you need to check things out from github" - are there documents to update?

kesara:
> There's a package philosophy change being made here from "a pip install gives you a dev environment" to "if you want a dev environment you need to check things out from github" - are there documents to update?

No, I think we should move away from `MANIFEST.in` and make the change to only use `pyproject.toml`.

Issue #1195, opened 2024-12-29
eref: internal link should be omitted in text mode


d-javu:
### Describe the issue

Linking to internal anchors: `[Link to somewhere](#anchor)` works very well for HTML output, where the text gets the link applied for ease of navigation.
In text output mode, however, the text `(#anchor)` gets sprinkled, and it is not a link, nor useful, in my opinion.
Should it be removed in text output?

### Code of Conduct

- [x] I agree to follow the [IETF's Code of Conduct](https://github.com/ietf-tools/.github/blob/main/CODE_OF_CONDUCT.md)

kesara:
Not an `xml2rfc` issue. See https://github.com/ietf-tools/xml2rfc/pull/1196#issuecomment-2576494334

PR #1196, opened 2024-12-30
text: do not render internal links in text mode

d-javu:
Avoids useless "... (#target) ..." link-like constructs in text output mode

Fixes bug #1195

This is on the level of 'It works for me' I don't know if something more sinister is necessary for this to be complete.

### DIFF
--- a/xml2rfc/writers/text.py
+++ b/xml2rfc/writers/text.py
@@ -1648,7 +1648,10 @@ class TextWriter(BaseV3Writer):
             self.warn(e, "Expected the 'target' attribute to have a value, but found %s" % (etree.tostring(e), ))
         if   brackets == 'none':
             if e.text and target:
-                target = "(%s)" % target
+                if target[0] == '#':
+                    target = ""
+                else:
+                    target = "(%s)" % target
         elif brackets == 'angle':
             target = "<%s>" % target                
         else:


rjsparks:
Hi Daniel - do you have a draft you're working on that shows the problem you're running into?

rjsparks:
It looks like you're trying to use `eref` for an internal link? Have you looked at `xref` (https://authors.ietf.org/en/rfcxml-vocabulary#xref)? 

If I'm misunderstanding the problem you're trying to solve, please elaborate?

d-javu:
See e.g. Section 4.4. of https://www.ietf.org/archive/id/draft-ietf-dmarc-dmarcbis-38.txt

```
4.4.  Identifier Alignment Explained

   DMARC validates the authorized use of the Author Domain (#author-
   domain) by requiring either that it have the same Organizational
   Domain (#organizational-domain) as an Authenticated Identifier
   (#authenticated-identifier) (a condition known as "Relaxed Alignment
   (#relaxed-alignment)") or that it be identical to the Authenticated
   Identifier (a condition known as "Strict Alignment (#strict-
   alignment)").  The choice of relaxed or strict alignment is left to
   the Domain Owner (#domain-owner) and is expressed in the domain's
   DMARC Policy Record (#dmarc-policy-record).  In practice, nearly all
   Domain Owners have found relaxed alignment sufficient to meet their
   needs.  Domain name comparisons in this context are case-insensitive,
   per [RFC4343].
```
All the `(#internal-link)` really is detrimental to the reading experience.

Another thing, not mentioned above, is mailto, and other links without a title,
who get an extra parenthesized copy of itself in text mode.

We are using mmark, and:
```
*  Aggregate feedback reports are sent via email to the address
   "dmarc-feedback@example.com"
   ("rua=mailto:dmarc-feedback@example.com")
```
is rendered as:
``` 
  *  Aggregate feedback reports are sent via email to the address
      "dmarc-feedback@example.com" ("rua=mailto:dmarc-
      feedback@example.com" (mailto:dmarc-feedback@example.com"))
```
unless you put "`" around it, like so
```
`("rua=mailto:dmarc-feedback@example.com")`
```
then the extra
```
(mailto:dmarc-feedback@example.com")
```
disappears.

See e.g B.2.3 of the older version https://www.ietf.org/archive/id/draft-ietf-dmarc-dmarcbis-30.txt

Note: here the correct thing is probably to quote with "`", because we want to avoid the link even in HTML mode.

To solve that problem is a few extra lines in the same area as the commit touches.

I doubt it is meant to be like that, but wanted to raise the issue to get some feedback.
I can add another commit to the PR if you agree this is an issue that should be fixed.

d-javu:
> It looks like you're trying to use `eref` for an internal link? Have you looked at `xref` (https://authors.ietf.org/en/rfcxml-vocabulary#xref)?
> 
> If I'm misunderstanding the problem you're trying to solve, please elaborate?

Given your comment this might be an mmark issue, but I am not well versed in it, not xml2rfc.

kesara:
I'm closing this PR since this is not a `xml2rfc` bug.

kesara:
I'm closing this PR since this is not a xml2rfc bug.

d-javu:
Thanks you for taking the time to give a comprehensive explanation,
that was very useful. I didn't know about the use title trick.
Looks like I need to read the syntax document more carefully.

PR #1225, opened 2025-02-14
feat: Remove warning about SVG width and height

martinthomson:
This warning annoys a lot of people unnecessarily.

The argument that this means that the diagram doesn't stretch is true (it won't stretch), but misleading, because stretching is not a good goal to seek.  Diagrams in RFCs often have text, which either becomes too large or too small when stretched.  Consequently, the right choice is to set width and height in many cases.

Indeed, the example in RFC 7992 includes width and height (and no viewbox, but that's a separate issue).  RFC 7996 does nothing about this.  This warning is therefore overreach.

### DIFF
--- a/xml2rfc/writers/html.py
+++ b/xml2rfc/writers/html.py
@@ -916,8 +916,6 @@ class HtmlWriter(BaseV3Writer):
             vbox = svg.get('viewBox')
             svgw = maybefloat(svg.get('width'))
             svgh = maybefloat(svg.get('height'))
-            if svgw or svgh:
-                self.warn(x, "Found SVG with width or height specified, which will make the artwork not scale.  Specify a viewBox only to let the artwork scale.")
             try:
                 if vbox:
                     xo,yo,w,h = re.split(',? +', vbox.strip('()'))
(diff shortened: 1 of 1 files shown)


kesara (approved):

cabo:
Yes, one of the warnings I regularly silence via:

```
v3xml2rfc:
  silence:
  - Found SVG with width or height specified
```

ajeanmahoney:
The RPC has struggled with this warning. Removing height and width attributes requires twiddling with the viewbox attribute to get the image to look sorta right, and this is very much a trial-and-error activity. Even if height and width are set, a reader can still make the image smaller or larger by using the zoom tools in the browser, so it's not like these attributes actually prevent scaling.

martinthomson:
That only highlights to me that getting SVG right is tricky (and perhaps in need of clearer guidance).  Removing width and height is not going to work in many cases.  For instance, I'm not aware of any cases with aasvg where you would want to do that rather than do something else entirely.  As Jean notes, zoom tools are probably a better fit from a usability perspective.

xml2rfc/writers/html.py
```diff
@@ -926,12 +924,11 @@ def render_artwork(self, h, x):
                     if not (svgw and svgh):
                         svgw = float(w)-float(xo)
                         svgh = float(h)-float(yo)
+                elif svgw and svgh:
+                    svg.set('viewBox', '0 0 %s %s' % (svgw, svgh))
```
rjsparks, line 928: I realize this is a pre-existing concern, but is overwriting any viewBox that might have been specified in the source the right thing to do?

Getting a good understanding of reasonable ways to use viewport and viewbox in the series I think needs some discussion.

rjsparks (approved):

xml2rfc/writers/html.py
ajeanmahoney: The accessibility experts recommend the following for understanding viewport and viewBox: https://www.sarasoueidan.com/blog/svg-coordinate-systems/
There's also https://stackoverflow.com/questions/15335926/how-to-use-the-svg-viewbox-attribute

PR #1237, opened 2025-05-10
fix: Local PIs need to override global ones, not the other way around

cabo:
`get_relevant_pis` searches from local to global, but then uses the most global PI, so a local override doesn't work.


cabo:
> @cabo, PR LGTM. I've added a test for this: [8948861](https://github.com/ietf-tools/xml2rfc/commit/8948861bf4fa370d86e0cde02e45810bf35b62c7)

Thank you!
Can you add a test for two sibling tables that each have a PI to each get the closest preceding PI?

cabo:
Thank you!

rjsparks (approved):

PR #1240, opened 2025-05-13
chore: Remove v3compat files

kesara:
These two files seems to be stale compared to `v3.rnc` & `v3.rng`.
I can't find any references to these files in the `xml2rfc` sourcecode.


rjsparks:
These were meant to be resources for other programs - the commit message for them read
```
A DTD approximation to the v3 schema.  It omits sschema for SVG and does not permit structured content (lists, figures, artwork, etc.) inside blockquotes, table cells, and list elements, as the full RelaxNG for this cannot be translated to the DTD format. - Legacy-Id: 3077


```
If they aren't being accessed or maintained, we should remove them (they probably should live somewhere else if we want to keep them anyhow).

rjsparks (approved):

Issue #1204, opened 2025-01-17
Ability to handle names that consist of a single letter


ajeanmahoney:
### Description

An author has requested that her name not be displayed with a period after it in the RFC header. Her full name is "Q Misell". Currently xml2rfc places a period after her first name because it creates an initial from ```<author fullname="Q Misell" role="editor" surname="Misell">```:

Current:
```
Automated Certificate Management Environment              Q. Misell, Ed.
Internet-Draft                                                  AS207960
Intended status: Standards Track                         14 January 2025
Expires: 18 July 2025
```
Requested:
```
Automated Certificate Management Environment               Q Misell, Ed.
Internet-Draft                                                  AS207960
Intended status: Standards Track                         14 January 2025
Expires: 18 July 2025
```

Something to consider, however, is what a code change to support this would do with author entries where a period is not provided in the initials attribute because xml2rfc adds them automatically. 

### Code of Conduct

- [x] I agree to follow the [IETF's Code of Conduct](https://github.com/ietf-tools/.github/blob/main/CODE_OF_CONDUCT.md)

kesara:
I didn't find any RFC specifying the author name format. The examples in RFC 7841 has the format: `initial. surname`. [^1]

Another option with current xml2rfc, is to just display the surname.
This can be done by adding `initials=""`.

[^1]: https://www.rfc-editor.org/rfc/rfc7841#appendix-A.1

cabo:
> I didn't find any RFC specifying the author name format.

That would be RFC 7322, Section 4.12. for "Author's Address" or "Authors' Addresses" Section, and Section 4.1.1 for the front page.

rjsparks:
Add this to the collection of evidence that we need to stop trying to disect names and treat them as blobs.

jrlevine:
Yup. Single letter first names aren't that unusual, e.g. former UN General Secretary U Thant

I know someone whose last name is O, no punctuation, just O

mferguson-rpc:
Just a note that this issue affects RFC-to-be 9799, which just entered AUTH48 state.

kesara:
@mferguson-rpc, Is the workaround I mentioned in https://github.com/ietf-tools/xml2rfc/issues/1204#issuecomment-2603449999 acceptable for this RFC-to-be?

For example:
```
    <author fullname="Q Misell" initials="" role="editor" surname="Misell">
```

will generate:
```
Internet Engineering Task Force (IETF)                       Misell, Ed.
```
[...]

PR #1246, opened 2025-06-04
Fix for #1204 - single letter names 

TheEnbyperor:
This PR updates the logic that generates short author names such that when initials are not provided, and the name being initialed is only one character long, there is no period appended after the character. 

This addresses the problem raised in #1204 regarding the handling of my name. The output format for RFC-to-be 9799 is now:
```text
Internet Engineering Task Force (IETF)                     Q Misell, Ed.
Request for Comments: 9799                                      AS207960
Category: Standards Track                                      June 2025
ISSN: 2070-1721
```

### DIFF
--- a/xml2rfc/util/name.py
+++ b/xml2rfc/util/name.py
@@ -37,7 +37,7 @@ def short_author_name_parts(a):
             if is_script(fullname, 'Latin'):
                 if len(fullname.split())>1:
                     parts = fullname.split()
-                    initials = ' '.join([ "%s."%n[0].upper() for n in parts[:-1] ])
+                    initials = ' '.join([ "%s."%n[0].upper() if len(n) > 1 else n.upper() for n in parts[:-1] ])
                     surname  = parts[-1]
                     parts = [initials, surname ]
                 else:
@@ -63,7 +63,7 @@ def short_author_ascii_name_parts(a):
         if fullname:
             if len(fullname.split())>1:
                 parts = fullname.split()
-                initials = ' '.join([ "%s."%n[0].upper() for n in parts[:-1] ])
+                initials = ' '.join([ "%s."%n[0].upper() if len(n) > 1 else n.upper() for n in parts[:-1] ])
                 surname  = parts[-1]
                 parts = [ initials, surname ]
             else:


ajeanmahoney:
While this presentation would be an acceptable variation of an author's name provided in the RFC header, I have concerns about how xml2rfc would handle the creation of reference entries. Currently it adds periods when a period doesn't exist in the fullname or initials attribute of the author element because sometimes writers can miss these.

For example, the following reference from RFC 8650 (note the lack of periods in fullname="I Hickson" and initials="I"):
```
        <reference anchor="W3C-20150203" target="https://www.w3.org/TR/2015/REC-eventsource-20150203/" quoteTitle="true" derivedAnchor="W3C-20150203">
          <front>
            <title>Server-Sent Events</title>
            <author fullname="I Hickson" initials="I" surname="Hickson">
              <organization showOnFrontPage="true"/>
            </author>
            <date day="03" month="February" year="2015"/>
          </front>
          <seriesInfo name="W3C" value="Recommendation"/>
          <annotation>Latest version available at &lt;<eref target="https://www.w3.org/TR/eventsource/"/>&gt;.</annotation>
        </reference>
```
Is rendered as "Hickson, I.,":
```
   [W3C-20150203]
              Hickson, I., "Server-Sent Events", W3C Recommendation, 3
              February 2015,
              <https://www.w3.org/TR/2015/REC-eventsource-20150203/>.
              Latest version available at <https://www.w3.org/TR/
              eventsource/>.
```

If an author wants their name to appear as "J. Doe" (with the period) in Authors' Addresses section, could they still do that with fullname="J. Doe"?

TheEnbyperor:
I have not touched the codepath if `initials` is set. Those will still always be rendered with a trailing period. So, either an author sets the initials field to "J", or puts the full name as "J. Doe".

kesara:
@ajeanmahoney This change affects references as well.
With this change:
```
<author fullname="J Doe" surname="Doe">
```

will generate:
```
Doe, J, "Title..
```

And,
```
<author fullname="J Doe" initials="J" surname="Doe">
```

will generate:
```
Doe, J., "Tile..
```

ajeanmahoney:
Would
```<author fullname="John Doe" surname="Doe">```
generate
```Doe, J, "Title...```
or
```Doe, J., "Title...```

We see instances of the following in non-RFC bibxml entries:
```
<author fullname="Christopher Meyer" surname="Meyer">
   <organization>Physical Measurement Laboratory</organization>
</author>
```

TheEnbyperor:
With my code, it would generate `Doe, J., "Title...`.

TheEnbyperor:
According to my tests, yes that's what it'll do.

Issue #1181, opened 2024-10-25
Get the full test suite on Windows OS


kesara:
### Description

#1177 currently allows only partial pytests (excluding some PDF checks) on Windows OS.
Current test suite should be updated or redesigned to work on Window OS.

### Code of Conduct

- [X] I agree to follow the [IETF's Code of Conduct](https://github.com/ietf-tools/.github/blob/main/CODE_OF_CONDUCT.md)

PR #1252, opened 2025-06-16
fix,test: Add missing test dependency

bkmgit:
python-dict2xml is needed for tests


pyproject.toml
```diff
@@ -46,7 +46,10 @@ documentation = "https://ietf-tools.github.io/xml2rfc/"
 issues = "https://github.com/ietf-tools/xml2rfc/issues"
 
 [project.optional-dependencies]
-pdf = ["weasyprint==65.0"]
+pdf = [
+    "weasyprint==65.0",
+    "dict2xml"
```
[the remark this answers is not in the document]
kesara: After that `tox.ini` can be configured install dependencies with `pip install ".[test]"`.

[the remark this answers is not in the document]
bkmgit: Thanks.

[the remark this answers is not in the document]
bkmgit: Hopefully ok now. Tests are run twice on MacOS:
- https://github.com/ietf-tools/xml2rfc/blob/main/.github/workflows/checks.yml#L167
- https://github.com/ietf-tools/xml2rfc/blob/main/.github/workflows/checks.yml#L176
and also on Linux:
- https://github.com/ietf-tools/xml2rfc/blob/main/.github/workflows/checks.yml#L109
- https://github.com/ietf-tools/xml2rfc/blob/main/.github/workflows/checks.yml#L115


Python dependency installation can also be removed from the checks.yml file for Linux
and MacOS other than tox, but not for Windows.

Makefile
```diff
@@ -66,10 +66,10 @@ env/bin/python:
 .PHONY: install
 install:
 	python3 --version
-	python3 -m pip install . --quiet
+	python3 -m pip install .[tests] --quiet
```
kesara, line 69: I think we should leave this as it is because `make install` can be used to more thank just running test. So we don't want to install test dependencies when we just want to build `xml2rfc`.

`python3 -m pip .[tests] --quiet` should go under the `test:` in line 72.

pyproject.toml
```diff
@@ -46,7 +46,19 @@ documentation = "https://ietf-tools.github.io/xml2rfc/"
 issues = "https://github.com/ietf-tools/xml2rfc/issues"
 
 [project.optional-dependencies]
-pdf = ["weasyprint==65.0"]
+pdf = [
+    "dict2xml",
```
kesara, line 50: `dict2xml` is a test dependency, so it should go under the `tests`.

pyproject.toml
```diff
@@ -46,7 +46,19 @@ documentation = "https://ietf-tools.github.io/xml2rfc/"
...
+    "weasyprint==65.0",
+]
+tests = [
+    "certifi",
+    "decorator",
+    "pyflakes",
+    "pypdf>=4.1.0",
+    "tox",
+    "tox-gh-actions",
+    "xml2rfc[pdf]",
```
kesara, line 60: Not really sure this will archive what we want. This might end up install `xml2rfc` from PyPI. I think we should add `"weasyprint==65.0"` as a separate dependency to tests.

kesara (changes requested):
@bkmgit, Thanks for the PR.
I have commented on couple of suggestions for further improvements.

pyproject.toml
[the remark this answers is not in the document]
kesara: Tests are run twice on purpose because first set of tests run with `yestests` which would updates the test outputs. 

Second run is done without `yestests` so second run will fail if there are any update to valid test outputs, as that should not happen.

Actually, thinking more about it first run might be improved by running the first set of tests on the git commit prior to the changes and next run on the post commits with changes. This is probably a separate PR.

The reason for Windows running tests just once is that we only run unit tests for Windows OS. See #1181.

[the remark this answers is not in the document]
bkmgit: Ok.

Makefile
bkmgit: Needed to create `installtestdeps` rather than just adding `python3 -m pip .[tests] --quiet`
Probably more of the commands in the Makefile should be indicated as .PHONY

pyproject.toml
```diff
@@ -46,7 +46,19 @@ documentation = "https://ietf-tools.github.io/xml2rfc/"
...
+]
+tests = [
+    "certifi",
+    "decorator",
+    "dict2xml",
+    "pyflakes",
+    "pypdf>=4.1.0",
+    "tox",
+    "tox-gh-actions",
+    "weasyprint==65.0",
```
kesara, line 60: Sorry for not spotting this earlier, This includes bunch of test dependencies that's only specific to github actions.
The list should only have:
```
    pyflakes
    decorator
    dict2xml
    pypdf>=4.1.0
    weasyprint==64.1
    tox
```

pyproject.toml
kesara: aha. `pip` dependency resolution seems to do the right thing. If you are keen I'm okay with going ahead with `"xml2rfc[pdf]",`

kesara (approved):

PR #1259, opened 2025-06-24
fix(pdf): Add generic mono-space font family

### LINKED ISSUE #1245: text-based "Artwork" in PDF misaligns depending on whether there is leading whitespace
### Describe the issue

See the attached XML file, which contains some text-based `<artwork>` objects.

[test.xml.txt](https://github.com/user-attachments/files/20527803/test.xml.txt)
 
textual artwork like this:
```
└┬╴Root of the tree
 ├─╴Child A
 └─╴Child B
```

Should show the vertical stem aligned, like so:

Image

with the `C` of the child rows under the first `o` of the root row.

In practice, in a PDF, it tends to be misaligned, like this:

Image

Interestingly, if there is a leading non-space character on each line, it seems to be correctly aligned:

Image

This is an impediment to the effective processing of [draft-ietf-lamps-e2e-mail-guidance](https://datatracker.ietf.org/doc/draft-ietf-lamps-e2e-mail-guidance/), which is in AUTH48 right now.

### Code of Conduct

kesara:
Addresses #1245

### DIFF
--- a/xml2rfc/writers/pdf.py
+++ b/xml2rfc/writers/pdf.py
@@ -136,7 +136,7 @@ class PdfWriter(BaseV3Writer):
         for script in scripts:
             family = get_noto_serif_family_for_script(script)
             fonts.add("%s" % family)
-        fonts = [ roboto_mono, ] + NOTO_SYMBOLS + list(fonts)
+        fonts = [ roboto_mono, ] + NOTO_SYMBOLS + list(fonts) + [ "monospace" ]
         self.note(None, "Found installed font: %s" % ', '.join(fonts))
         return fonts
 


kesara:
@dkg, my suspicion is Weasyprint is not choosing Roboto Mono font for the arrow characters.
The issue with all these different codepoints seems to be font related.

kesara:
In a docker instance that has Roboto Mono fonts from xml2rfc-fonts installed,
I didn't get Roboto Mono as a choice for following commands:
```
fc-list ':charset=21A7 :monospace'
fc-list ':charset=21E9 :monospace'
```

dkg:
Hm, OK, i don't think i have Roboto Mono installed locally, but pulling it from https://fonts.google.com/specimen/Roboto+Mono and looking at its `cmap` table using `ttfdump` suggests that you're right that U+21A7 and U+21E9 are not in the table.

In fact, i don't see any useful "downward arrows" glyphs in Roboto Mono.  Furthermore, there are no BOX DRAWINGS glyphs in Roboto Mono either (nothing in the U+25XX range at all, if i'm interpreting the `cmap` table correctly).

Using `pdf2txt -t xml rfc9787.pdf`, the BOX DRAWINGS glyphs are all supplied with DejaVu Sans Mono, and those glyphs are all 5.719 units wide, very close to the neighboring Noto Mono glyphs, which are 5.7 units wide:

```
<text font="IUHGEO+DejaVu-Sans-Mono" bbox="76.156,345.749,81.875,355.249" colourspace="DeviceRGB" ncolour="(0.133333, 0.133333, 0.133333)" size="9.500">└</text>
<text font="IUHGEO+DejaVu-Sans-Mono" bbox="81.875,345.749,87.594,355.249" colourspace="DeviceRGB" ncolour="(0.133333, 0.133333, 0.133333)" size="9.500">┬</text>
<text font="IUHGEO+DejaVu-Sans-Mono" bbox="87.594,345.749,93.313,355.249" colourspace="DeviceRGB" ncolour="(0.133333, 0.133333, 0.133333)" size="9.500">╴</text>
<text font="DPFAUG+Roboto-Mono" bbox="93.313,345.417,99.013,354.917" colourspace="DeviceRGB" ncolour="(0.133333, 0.133333, 0.133333)" size="9.500">m</text>
```

But the DOWNWARDS ARROW glyphs (and all preceding whitespace) are filled in with Noto Sans Symbols2, which isn't even a monospace font.  its space character is 2.47 units wide, and its "⇩" is 6.289, compared to the Noto Mono glyphs, which remain 5.7:

```
<text font="QOSDXS+Noto-Sans-Symbols2" bbox="76.156,214.847,78.626,224.347" colourspace="DeviceRGB" ncolour="(0.133333, 0.133333, 0.133333)" size="9.500"> </text>
<text font="QOSDXS+Noto-Sans-Symbols2" bbox="78.626,214.847,84.915,224.347" colourspace="DeviceRGB" ncolour="(0.133333, 0.133333, 0.133333)" size="9.500">⇩</text>
<text font="DPFAUG+Roboto-Mono" bbox="84.915,218.257,90.615,227.757" colourspace="DeviceRGB" ncolour="(0.133333, 0.133333, 0.133333)" size="9.500"> </text>
```

So i have a couple specific questions for how to resolve this:

- given that the BOX DRAWINGS and DOWNWARDS ARROWS glyphs are all in DejaVu Sans, why isn't xml2rfc selecting DejaVu Sans as the fallback font in the monospace context for the DOWNWARDS ARROWS, when it is successfully selecting it for the BOX DRAWINGS?
- even if we're using a fallback font for missing glyphs, shouldn't the leading whitespace at least be rendered in Roboto Mono, to keep the spacing for the line's initial offset correct?

dkg:
What if you put `+ [ "monospace" ]` before `NOTO_SYMBOLS` instead of after it, in this font list?  or what if you explicitly ask for `Noto Sans Mono` as a fallback for glyphs missing from `Roboto Mono`, as suggested in #1261

kesara:
Addressed by #1261

PR #1261, opened 2025-06-26
fix(pdf): For monospaced text, Add Noto Sans Mono before NOTO_SYMBOLS

dkg:
This is an attempt to address #1245

### DIFF
--- a/xml2rfc/writers/pdf.py
+++ b/xml2rfc/writers/pdf.py
@@ -133,10 +133,11 @@ class PdfWriter(BaseV3Writer):
         fonts = set()
         scripts = self.root.get('scripts').split(',')
         roboto_mono = "Roboto Mono"
+        noto_sans_mono = "Noto Sans Mono"
         for script in scripts:
             family = get_noto_serif_family_for_script(script)
             fonts.add("%s" % family)
-        fonts = [ roboto_mono, ] + NOTO_SYMBOLS + list(fonts)
+        fonts = [ roboto_mono, noto_sans_mono, ] + NOTO_SYMBOLS + list(fonts)
         self.note(None, "Found installed font: %s" % ', '.join(fonts))
         return fonts
 


dkg:
The idea here is to keep in the Roboto and Noto families, but to increase the range of covered glyphs in a fixed-width font.

I'm not sure how well this will work, because i don't think i have the same font environment that the RFC editor's toolchain uses when it works.  but hopefully this makes it easier to test, at least.

kesara:
This is a better solution and this change stops leaking non-RFC fonts.
But monospace arrows are not aligning well. See [pdf](https://t4.fq.nz/xml2rfc.1261/rfc9787.pdf) generated with this change.

Screenshot 2025-06-27 at 16 46 45
Screenshot 2025-06-27 at 16 46 54

kesara:
@dkg If you want to use the same fonts as RFC editor, you can get them from [xml2rfc-fonts](https://github.com/ietf-tools/xml2rfc-fonts) or you xml2rfc docker instace with `./run` from project root dir. Just make sure xml2rfc docker images are up-to-date.

xml2rfc/writers/pdf.py
```diff
@@ -133,10 +133,12 @@ def get_mono_fonts(self):
         fonts = set()
         scripts = self.root.get('scripts').split(',')
         roboto_mono = "Roboto Mono"
+        noto_sans_mono = "Noto Sans Mono"
+        dejavu_sans_mono = "DejaVu Sans Mono"
```
kesara, line 137: PDFs have been restricted to Noto and Roboto Mono fonts.
I guess this needs attention of rapt/rswg.

dkg:
I've dropped the commit from this series that included DejaVu Sans Mono, to focus specifically on Noto Sans Mono, as it should be sufficient for a BOX-DRAWINGS-only resolution to the specific issues.

Maybe this is safe to merge as-is now, regardless of the outcome of the rpat discussion?  I can always open up a separate MR for further discussion related to DejaVu Sans Mono.

PR #1282, opened 2025-09-24
feat(html): Use self hosted fonts

### LINKED ISSUE #1281: Use self hosted fonts instead of google fonts
### Description

At the moment RFC & I-D HTML uses google fonts[^1].

Now the required fonts are also available on ietf.org:
* https://static.ietf.org/fonts/noto-sans/import.css
* https://static.ietf.org/fonts/noto-serif/import.css
* https://static.ietf.org/fonts/roboto-mono/import.css
* https://static.ietf.org/fonts/noto-sans-math/import.css [^2]
* https://static.ietf.org/fonts/noto-sans-symbols-2/import.css [^2]

`xml2rfc` should use fonts these self hosted fonts instead of google fonts.

See the recent tools-discuss discussion[^3] about fonts as well.

[^1]: https://github.com/ietf-tools/xml2rfc/blob/main/xml2rfc/data/xml2rfc.css#L14-L17
[^2]: These fonts are not defined explicitly in current CSS but there are recent RFCs that make use of those.
[^3]: https://mailarchive.ietf.org/arch/msg/tools-discuss/4hvXbctvojOyRhXUxbwFS0yYKko/

### Code of Conduct

kesara:
Fixes #1281 

This PR updates CSS to use self hosted fonts.

### DIFF
--- a/tests/valid/docfile.html
+++ b/tests/valid/docfile.html
@@ -4,15 +4,15 @@
 <meta charset="utf-8">
 <meta content="Cherokee,Common,Greek,Latin" name="scripts">
 <meta content="initial-scale=1.0" name="viewport">
-<title>Xml2rfc Vocabulary Version 3 Schema xml2rfc release 3.30.0</title>
+<title>Xml2rfc Vocabulary Version 3 Schema xml2rfc release 3.30.2</title>
 <meta content="xml2rfc(1)" name="author">
 <meta content="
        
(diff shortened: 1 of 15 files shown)


kesara:
Example RFC generated with self hosted fonts: [RFC 9840](https://t4.fq.nz/xml2rfc-1282/rfc9840.html)

rjsparks (approved):

PR #1299, opened 2025-10-21
fix: Emit an error when there's no author

### LINKED ISSUE #1274: Wrong error message in xml2rfc if there is no author
### Describe the issue

When xml2rfc is parsing a &lt;front&gt;, if there is no &lt;author&gt; but there is a &lt;date&gt;, it reports that it did not expect a date in the element. It should instead report that there is no &lt;author&gt; but one is required.

### Code of Conduct

kesara:
Fixes #1274


kesara:
See https://github.com/ietf-tools/xml2rfc/issues/1274#issuecomment-3431162089

PR #1304, opened 2025-11-04
feat: new xml2rfc-slim docker image

daniel-dewa:
**Problem**

Many specs from Oidf use a custom (unmaintained) docker image containing [mmark](https://github.com/mmarkdown/mmark) and xml2rfc for CI / CD builds. mmark recently added an official docker image containing the latest and versioned releases of the cli tool for this use case.

The image `ghcr.io/ietf-tools/xml2rfc-base` exists, but since it is **2.28 GB** big, it is AFAIK unused because of its size.

**Solution**

My PR adds a new image, called `xml2rfc-slim`, which is meant for external usage, primarily CI / CD. The size is kept small by excluding pdf support and using a two-step Dockerfile. I hope the IETF Tools team will merge the new image and use the provided changes to the xml2rfc CI / CD.

I'm not an expert in GitHub actions and recommend you check `pypi-publish.yml` still works as expected.

### DIFF
--- a/README.md
+++ b/README.md
@@ -81,22 +81,31 @@ With these installed and available to **xml2rfc**, the `--pdf` switch will be en
 
 ### Docker container
 
-A cross-platform docker image is available (at `ghcr.io/ietf-tools/xml2rfc-slim:latest`) for CI / CD usage. The image does NOT contain `pdf` support, as to keep the image size small (**175 MB**).
+A series of docker images are available to accommodate various use cases. The table below describes the features of each image type. 
+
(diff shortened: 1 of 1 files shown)


kesara:
@daniel-dewa, Thanks for the PR.
For CI/CD purpose `xml2rfc` can be installed from PyPI with `pip` or similar tool.
So there's a light option available.
I'm not sure publishing a light docker image would be worth while.

Reason for publishing a large xml2rfc-base image is so that it includes fonts that required for PDF generation.

I'd like know your thoughts about this.

PR #1307, opened 2026-02-09
fix: Handle unblance table errors

### LINKED ISSUE #740: text rendering of table leads to crash
complex table renders ok in HTML and PDF, but crashes when making text.

When using ```<table>``` in XML:
- https://www.rfc-editor.org/v3test/draft-zia-route-table2.xml
- https://www.rfc-editor.org/v3test/draft-zia-route-table2.html 
- https://www.rfc-editor.org/v3test/draft-zia-route-table2.pdf 
(orig table: https://datatracker.ietf.org/doc/html/draft-zia-route-06#section-10)

TXT: crashes w/ Traceback as follows.
```Traceback (most recent call last):
  File "/usr/bin/xml2rfc", line 8, in <module>
    sys.exit(main())
  File "/usr/lib/python3.6/site-packages/xml2rfc/run.py", line 742, in main
    writer.write(filename)
  File "/usr/lib/python3.6/site-packages/xml2rfc/writers/text.py", line 343, in write
    text = self.process()
  File "/usr/lib/python3.6/site-packages/xml2rfc/writers/text.py", line 301, in process
    lines = self.render(self.root, width=72, joiners=joiners)
  File "/usr/lib/python3.6/site-packages/xml2rfc/writers/text.py", line 367, in render
    res = func(e, width, **kwargs)
  File "/usr/lib/python3.6/site-packages/xml2rfc/writers/text.py", line 3054, in render_rfc
    lines = self.ljoin(lines, c, width, **kwargs)
  File "/usr/lib/python3.6/site-packages/xml2rfc/writers/text.py", line 554, in ljoin
    res = mklines(self.render(e, width, **kwargs), e)
  File "/usr/lib/python3.6/site-packages/xml2rfc/writers/text.py", line 367, in render
    res = func(e, width, **kwargs)
  File "/usr/lib/python3.6/site-packages/xml2rfc/writers/text.py", line 2232, in render_middle
    lines = self.ljoin(lines, c, width, **kwargs)
  File "/usr/lib/python3.6/site-packages/xml2rfc/writers/text.py", line 554, in ljoin
    res = mklines(self.render(e, width, **kwargs), e)
  File "/usr/lib/python3.6/site-packages/xml2rfc/writers/text.py", line 367, in render
    res = func(e, width, **kwargs)
  File "/usr/lib/python3.6/site-packages/xml2rfc/writers/text.py", line 3325, in render_section
    lines = self.ljoin(lines, c, width, **kwargs)
  File "/usr/lib/python3.6/site-packages/xml2rfc/writers/text.py", line 554, in ljoin
    res = mklines(self.render(e, width, **kwargs), e)
  File "/usr/lib/python3.6/site-packages/xml2rfc/writers/text.py", line 367, in render
    res = func(e, width, **kwargs)
  File "/usr/lib/python3.6/site-packages/xml2rfc/writers/text.py", line 4250, in render_table
    lines = self.build_table(e, width, **kwargs)
  File "/usr/lib/python3.6/site-packages/xml2rfc/writers/text.py", line 4143, in build_table
    k, l = cell.origin
AttributeError: 'Cell' object has no attribute 'origin'
```

kesara:
Cherry picked changes from #801.

Fixes #740

### DIFF
--- a/xml2rfc/writers/text.py
+++ b/xml2rfc/writers/text.py
@@ -4170,18 +4170,24 @@ class TextWriter(BaseV3Writer):
             minspan = sys.maxsize
             for j in range(cols):
                 cell = cells[i][j]
-                k, l = cell.origin
-                hspan = cell.rowspan+k-i if cell.rowspan else minspan
-                if hspan > 0 and hspan < minspan:
-                    minspan = hspan
+                if hasattr(cell, 'origin'):
+                    k, l = cell.origin
+                    hspan = cell.rowspan+k-i if cell.rowspan else minspan
+                    if hspan > 0 and hspan < minspan:
+                        minspan = hspan
+                else:
+                    self.die(e, "Inconsistent table width: Found different row lengths in this table")
             maxlines = 0
             for j in range(cols):
                 cell = cells[i][j]
-                k, l = cell.origin
-                hspan = cell.rowspan+k-i if cell.rowspan else minspan
-                lines = len(cell.wrapped) if cell.wrapped else 0
-                if lines > maxlines:
-                    maxlines = lines
+                if hasattr(cell, 'origin'):
+                    k, l = cell.origin
+                    hspan = cell.rowspan+k-i if cell.rowspan else minspan
+                    lines = len(cell.wrapped) if cell.wrapped else 0
+                    if lines > maxlines:
+                        maxlines = lines
+                else:
+                    self.die(e, "Inconsistent table width: Found different row lengths in this table")
             for j in range(cols):
                 cells[i][j].lines = maxlines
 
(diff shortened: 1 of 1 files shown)


jennifer-richards (commented):
I don't think this _fixes_ the issue. It's an improvement, but it leads to divergent behavior between output formats. To _fix_ the initial issue, I think `xml2rfc` either needs to reject the document in all cases (but that may run into Carsten's comment on the original PR about it possibly being valid to have unbalanced tables...) or render the text version in a way equivalent to what PDF/HTML do.

No objection to merging this as an incremental improvement if you want to remove the "fixes" assertion.

holloway:
It's hard for me to comment on this PR specifically, so just some general questions/comments.

I know this isn't particularly helpful feedback for this PR specifically but I have to start somewhere.

* for context, we take into RFC XML and generate ascii.. and this is the ascii table generator? Seems a very odd thing to maintain ourselves. It seems that we should generate HTML and use a conventional html2ascii converter rather than maintaining our own (ascii layout is _very_ complex, we should use a mature library). The pattern of 'html -> output format' is already used for PDF. Has this been considered? What bespoke functionality needs to exist in the ascii output?
* it seems we should have tests for input and output transformations like this. I don't see snapshot tests in the PR. It's hard to tell whether this change regresses the rendering of other things.

kesara:
> It's hard for me to comment on this PR specifically, so just some general questions/comments.
> 
> I know this isn't particularly helpful feedback for this PR specifically but I have to start somewhere.
> 
>     * for context, we take into RFC XML and generate ascii.. and this is the ascii table generator? Seems a very odd thing to maintain ourselves. It seems that we should generate HTML and use a conventional html2ascii converter rather than maintaining our own (ascii layout is _very_ complex, we should use a mature library). The pattern of 'html -> output format' is already used for PDF. Has this been considered? What bespoke functionality needs to exist in the ascii output?
> 

It would be nice to use a library but yet to find a table generation library that can fulfill what we need. I think I've explored [prettytable](https://github.com/prettytable/prettytable) and [texttable](https://github.com/foutaise/texttable/) before they haven' had the feature we want. What you are proposing with `html2ascii` is major change for the project. Which might be viable if HTML output can output ASCII output similar to what published TXT.

>     * it seems we should have tests for input and output transformations like this. I don't see snapshot tests in the PR. It's hard to tell whether this change regresses the rendering of other things.

There's no rendering output since this was designed to generate an error.
Maybe there's room for adding tests for the table function.

kesara:
> I don't think this _fixes_ the issue. It's an improvement, but it leads to divergent behavior between output formats. To _fix_ the initial issue, I think `xml2rfc` either needs to reject the document in all cases (but that may run into Carsten's comment on the original PR about it possibly being valid to have unbalanced tables...) or render the text version in a way equivalent to what PDF/HTML do.
> 
> No objection to merging this as an incremental improvement if you want to remove the "fixes" assertion.

Yeah, I think there should be validation at the prep step so HTML/PDF would reject and warn authors about unbalanced table.

jennifer-richards:
> Yeah, I think there should be validation at the prep step so HTML/PDF would reject and warn authors about unbalanced table.

I can't see the examples that were linked in the issue - the links are broken - but [this comment](https://github.com/ietf-tools/xml2rfc/issues/740#issuecomment-1162835378) makes me a bit wary about declaring the imbalance an error.

rjsparks:
@holloway  - the ascii table generator has been hand-tuned for our particular audience and text format rules for many years - a switch to render through html to ascii would likely bring a large wave of issues to address mismatches that are already addressed. If we were to attempt making that architectural change to the code, we would want to run parallel tests over a significant part of the corpus and reconcile differences.

PR #1322, opened 2026-05-14
feat(html): add copy unfolded button for RFC 8792 folded blocks

panva:
Detect [RFC 8792](https://www.rfc-editor.org/rfc/rfc8792.html) folded blocks in the v3 HTML renderer and add a hover/focus copy control that copies the block text unfolded.

This makes RFC 8792-folded examples usable directly from HTML RFCs: readers can copy the original unfolded source without needing to know or manually apply the folding rules.

rfc8792-sourcecode-hover
rfc8792-sourcecode-copied

Note that I am in no way shape or form a frontend stylesheet proficient, nor knowledgeable of aria-* accessibility attributes. The style, text, and attributes need some bashing and I'm happy to accomodate any requested changes.

### DIFF
--- a/tests/input/sourcecode.xml
+++ b/tests/input/sourcecode.xml
@@ -127,6 +127,35 @@ print("49")
 print("50")
             </sourcecode>
         </section>
+        <section>
+            <name>RFC 8792 wrapping</name>
+            <t>Source code using the single backslash strategy:</t>
+            <sourcecode name="single-wrap.txt"><![CDATA[========== NOTE: '\' line wrapping per RFC 8792 ===========
+
+first folded line \
+continues here
+second folded line \
+also continues
+]]></sourcecode>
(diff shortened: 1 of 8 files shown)


bc-pi:
✅

kesara:
@panva, Your PR proposes changes _to_ one of the published _formats_ of the RFCs. I'll add this to next [RPAT](https://datatracker.ietf.org/group/rpat/about/) meeting to discuss how to proceed with this.

The new revamped [www.rfc-editor.org](https://www.rfc-editor.org/) has a different HTML presentation of RFCs. For example: https://www.rfc-editor.org/info/rfc8792/

May be this change is better suited for the new RFC editor website?

New RFC editor repository is https://github.com/ietf-tools/red/

kesara:
@panva Thanks for the PR. I'm closing this as similar feature has been implemented in rfc-editor website.

PR #1330, opened 2026-06-08
fix(html): Update xml2rfc.css with sub/sup no wrapping

holloway:
Apparently `<sub>`/`<sup>` shouldn't wrap, according to RPC.

I'd guess that most usage only has short text so this is probably ok to do.

Let me know if you'd accept the PR in principle and then I'll make the PR pass


kesara:
Looks like the right thing to do.

kesara (approved):

PR #1338, opened 2026-07-22
fix(pdf): Improve logging

### LINKED ISSUE #1337: Suspress fontTools.varLib.instancer info logs
### Describe the issue

Looks like `xml2rfc` is emitting `fontTools.varLib.instancer` log messages.
They might be a bug in  xml2rfc logic for capturing the `fontTools.varLib` logger[^1].

[^1]: https://github.com/ietf-tools/xml2rfc/blob/main/xml2rfc/writers/pdf.py#L49C43-L49C52

### Code of Conduct

kesara:
Fixes #1337

### DIFF
--- a/xml2rfc/writers/pdf.py
+++ b/xml2rfc/writers/pdf.py
@@ -44,9 +44,7 @@ class PdfWriter(BaseV3Writer):
 
         loggers = []
         loggers.append(logging.getLogger("weasyprint"))
-        loggers.append(logging.getLogger("fontTools.subset"))
-        loggers.append(logging.getLogger("fontTools.ttLib"))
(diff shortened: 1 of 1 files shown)


rudimatz (approved):

jennifer-richards (commented):
code change looks ok, but I can't confirm that the behavior it'll induce is "correct" :-)

kesara:
> code change looks ok, but I can't confirm that the behavior it'll induce is "correct" :-)

IIRC Reason for having multiple logger entries was general logger entry didn't catch them all before. This issue seem to have corrected in fonttools package.

holloway (approved):

PR #1345, opened 2026-09-10
build: Add PDF/A-3U verification workflow

### LINKED ISSUE #1344: Add a GHA to verify the PDF/A validity
### Description

Since [v3.33.0](https://github.com/ietf-tools/xml2rfc/releases/tag/v3.33.0) `xml2rfc` is generating PDF documents with PDF/A-3U profile.
There should be a GHA to test PDF/A-3U validity.

### Code of Conduct

kesara:
Add a GitHub Actions workflow that generates a PDF from a sample RFC input and verifies PDF/A-3U conformance using veraPDF, matching the pdf/a-3u variant produced by WeasyPrint.

Fixes #1344

Model: claude-opus-4-8 (thinking: medium)

### DIFF
--- a/.github/workflows/pdfa-verification.yml
+++ b/.github/workflows/pdfa-verification.yml
@@ -85,9 +85,8 @@ jobs:
       run: |
         mkdir -p tests/out
         xml2rfc --skip-config --allow-local-file-access --no-network \
-          --pdf --v3 --legacy-date-format \
+          --pdf --attach-xml --v3 \
           --rfc-reference-base-url https://rfc-editor.org/rfc \
-          --id-reference-base-url https://datatracker.ietf.org/doc/html/ \
           tests/input/rfc9001.canonical.xml --out tests/out/rfc9001.pdf
 
     - name: Verify PDF/A-3U conformance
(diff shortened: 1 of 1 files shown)


rjsparks (approved):
Approving. Note, however, the maintenance issue at L36. Fonts don't change often, but it might be hard to remember to bump this.

kesara:
> Approving. Note, however, the maintenance issue at L36. Fonts don't change often, but it might be hard to remember to bump this.

I agree, we might want to find a way to configure fonts version in a single place. But I'm happy to take the risk for now.

jennifer-richards (approved):
Looks reasonable. I think it might be preferable long-term to build a docker image for running the tests rather than downloading and installing verapdf etc every time. That would also give a hook for the maintenance that Robert pointed at.

kesara:
I created https://github.com/ietf-tools/xml2rfc/issues/1346 to capture font configuration maintenance burden.

kesara:
> Looks reasonable. I think it might be preferable long-term to build a docker image for running the tests rather than downloading and installing verapdf etc every time. That would also give a hook for the maintenance that Robert pointed at.

May be this could be part of dev docker image.

